PR #106, opened 2017-06-29
Fixed for loop to correctly pad end of files

ibremseth:
I noticed the padding did not always complete the block of 10 records. It looked like the for loop was not looping the correct amount of time. I fixed the loop condition. 

Other then that this works amazing! 
@wadearnold


wadearnold:
Thank you Isaac for the pull request!Please let me know if I can help improve it for your use cases. I'm pretty specific to PPD files for my own usage.

Issue #105, opened 2017-04-20
Strategy for Adding Support for More Batch Types


jjafuller:
I would like to help out adding more batch types, but based on the way the library is currently written I am not sure how you were planning on adding support for batch types other than PPD.

Can you point me in the right direction?

wadearnold:
I don't have a hashed out strategy but do have some ideas. 

I was going to make a Batch interface w/ parse, validate, write, build functions exposed along with any EntryDetail and Addend that was specific to the batch type. Right now these are explicit for PPD files and need to change to (SEC)Header, (SEC)EntryDetail, (SEC)Addendum, (SEC)Control for each of the matching SEC (Standard entry class code) of the different batch types. 

The file reader would then need to look at the SEC code of the Batch Header and then use a switch based on the implemented SEC codes. Unsupported SEC codes would still default to an unsupported error. 

When building a new batch file the NewBatch function will need to take a SEC code.

wadearnold:
@jjafuller  Starting to try and abstract the Batch SEC type into an interface and make the Batch names,  Entry Details, and validation explicit to the SEC code. 
https://github.com/moov-io/ach/compare/master...wadearnold:batch-sec-interface-refactor

jjafuller:
It sounds like you're on the right track. One thing to consider is that many of the SEC batch record formats are the same, so the trick is achieving the correct functionality without duplicating logic more than necessary. 

Do you have a priority on which SEC codes you want to add? I know that ones I work with the most frequently are PPD, CCD, WEB, and TEL. However, I do work with ENRs, and IAT fairly regularly. IATs are always an issue since they play by their own rules.

wadearnold:
What I am thinking is that:

1. Make the Reader, Writer, and all test use the batcher interface. Only BatchPPD concrete type implements the batcher interface at this time. This refactor should get the code back to its current working state for PPD files. 

2. Extract the common Batch, Entry Detail, and Addendum functionality into concrete types that are not used by themselves but composed into the SEC code type. I assume this to be highly iterative per type to figure out what is repeated so that we have DRY code. 
`type batchRCK  struct {
      Batch
      // code specific to RCK implementation
 }`

3. Continue implementing the different SEC codes while refactoring out common code to the composable struct for each SEC type. 

I know this is pretty high level but what are your thoughts @jjafuller ?

wadearnold:
[...]

PR #129, opened 2017-08-23
Composable batch

wadearnold:
Changes to make adding additional payment types via SEC codes easier to support.


wadearnold:
Adds support for a composable batch struct into SEC specific batch types #105

PR #141, opened 2017-10-17
Implement COR SEC   Return File Processing

wadearnold:
Thanks @killianbrackey and @Sezzle for all the work in adding support.


wadearnold:
The deeper I jump into this rabbit hole the more I think we need a Addenda, AddendaReturn, AddendaCOR, AddendaIAT, etc like we have done with Batch. Just talking to myself publicly.

wadearnold:
Adding changes to master that support @zelle changes. I am working on a new branch that address the Addenda types such as NOC, IAT, etc that need to eventually be supported. @killianbrackey you may need to comment out the ReturnAddenda validator on the ReturnCodeDict so that this works for BatchCOR/NOC until I get the second phase of this refactor committed.

PR #143, opened 2017-10-30
Addendumer addenda 

wadearnold:
Create support for BatchCOR types and expose them file.NotificationOFChange


wadearnold:
This PR has a breaking change to the API. Previously NewAddenda() returned 1 variable. 

```go
addenda := ach.NewAddenda()
```

Now it returns 2 variables with the second being an error if it exists. 

```go
addenda, err  := ach.NewAddenda()
if err != nil {
   // do something with the error
}
```

Issue #169, opened 2018-04-19
Utilize Go 1.10 strings.Builder & strings.Grow


wadearnold:
Go 1.10 introduced the [strings.Builder](https://golang.org/pkg/strings/#Builder) type is used to efficiently build a string using write methods.It offers a subset of the [bytes.Buffer](https://golang.org/pkg/bytes/#Buffer) methods that allows it to safely avoid redundant copying. The [strings.Grow](https://golang.org/pkg/strings/#Builder.Grow) method can be used to preallocate memory when the maximum size of the string is known. 

Utilizing these new language features it may be possible to allocate 94 characters in each of the records String() struct and improve the write performance of the library.

PR #237, opened 2018-08-15
switch String() to use strings.Builder internally

adamdecaf:
I'm not actually sure this helps a lot. We're still creating a lot of internal strings (i.e. from alphaField` and `numericField`). I ran a couple benchmarks locally and didn't see allocs decrease, and the ns/op time didn't change much. 

Fixes: https://github.com/moov-io/ach/issues/169

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: go
 sudo: false
 go:
-- 1.9.x
 - 1.10.x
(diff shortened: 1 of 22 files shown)


adamdecaf:
Note: This also now requires Go 1.10+, and updates the CI job accordingly.

PR #247, opened 2018-08-18
test: add Fuzz support for Reader

adamdecaf:
Fuzzing is a technique for sending arbitrary input into functions to see what happens. Typically this is done to weed out crashes/panics from software, or to detect bugs. In some cases all possible inputs are ran into the program (i.e. if a function accepts 16-bit integers it's trivial to try them all).

In this PR we are fuzzing `reader.go` from the root level -- In specific the `Reader.Read()` method.


adamdecaf:
There isn't really a good CI story for fuzzing yet, but running this overnight or a few days would be a good start.

.gitignore
```diff
@@ -39,3 +39,9 @@ _testmain.go
 # code coverage
 coverage.html
 cover.out
+
+# fuzzing
+test/fuzz-reader/corpus
+test/fuzz-reader/crashes
+test/fuzz-reader/suppressions
+test/fuzz-reader/*.zip
```
adamdecaf, line 50: These files just bloat the repo. We can share them between ourselves (or from the community) if needed. Keeping them around helps later runs too.

wadearnold:
Do you want me to leave this for open or merge it in @adamdecaf

adamdecaf:
I want to run this overnight and see if anything comes about. Setting up my new laptop to do that though.

adamdecaf:
I've ran this a bit locally, but I think the corpus needs built better (so that we generate partial files).

adamdecaf:
_re-reads the go-fuzz docs_

I should have generated a few example files and loaded those into the corpus..

adamdecaf:
Examples are in, but Go 1.11 modules break `go-fuzz`.

wadearnold (approved):

adamdecaf:
Two crashes found via `go-fuzz` are fixed, with regression test cases added. There's a lot of index based access into strings, which panics if the strings aren't long enough. 

We might want to add a string length validator. Thoughts @bkmoovio ? The same likely applies to X9.

wadearnold (approved):

Issue #254, opened 2018-09-04
Supply a REST interface for file creation


adamdecaf:
To further adoption we want to generate a REST (HTTP) interface for the library. Currently you have to know ACH and GoLang to leverage the library, but this would allow people to just understand ACH and use a library in a language they know. 

Required:
- [ ] good test coverage
- [x] structured logging
- [x] go-kit with metrics, health checks
- [ ] sanity specs (i.e. build file, add batches, render)
- [x] `/metrics` endpoint on admin servlet 
   - Using https://github.com/heptiolabs/healthcheck ? 
- `/pprof` on admin servlet

There's also a [swagger api spec (draft)](https://github.com/wadearnold/achsvc/blob/master/api/openapi.yaml) that will be used to generate clients from. 

Releases should contain: 
 - [ ] pre-built binaries (on Github release)
 - [ ] tagged docker image on docker hub

Note: [genuinetools/img](https://github.com/genuinetools/img/blob/master/Makefile) as an example of making releases

 We are working on making the endpoint a direct mapping to the library. Something like this: 

```
    // POST   /files/                          Creates a file
    // GET    /files/                          retrieves a list of all file's
    // GET    /files/:id                       retrieves the given file by id
    // DELETE /files/:id                       delete a file based on supplied id
    // GET    /files/:id/validate          validate records    
    // GET    /files/:id/contents         assemble file (totals, sequence numbers, etc) and return plaintext form

    // POST   /files/:id/batches/              Create a Batch
    // GET    /files/:fileID/batches           Retrieve a list of all batches for the file id.
    // GET    /files/:id/batches/:id           Retrieve the given batch by id
    // DELETE /files/:fileID/batches/:batchID  delete the batch based on the supplied file and batch id
```

adamdecaf:
This is done. Adding tests is tracked in other issues.

PR #255, opened 2018-09-04
REST endpoints for file creation 

adamdecaf:
Creates a REST interface for building ACH files. This is also going to be our first deployed service. 

Issue: https://github.com/moov-io/ach/issues/254


cmd/server/main.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/wadearnold/ach/server"
+	"github.com/wadearnold/kit/log"
```
adamdecaf: fix import path

server/repositoryInMemory.go
```diff
@@ -0,0 +1,123 @@
...
+	FindFile(id string) (*ach.File, error)
+	FindAllFiles() []*ach.File
+	DeleteFile(id string) error
+	StoreBatch(fileID string, batch ach.Batcher) error
+	FindBatch(fileID string, batchID string) (*ach.Batcher, error)
+	FindAllBatches(fileID string) []*ach.Batcher
+	DeleteBatch(fileID string, batchID string) error
+}
+type repositoryInMemory struct {
+	mtx   sync.RWMutex
```
adamdecaf: @wadearnold a bit of a style thing, but I sometimes embed right into the struct (but you expose Lock/Unlock publically here. Since `repositoryInMemory` isn't exported though we don't have to worry as much about letting loose implementation details.

```Go
type Foo struct {
  sync.Mutex
}

// func (f *Foo) Action() error { // calls f.Lock() / f.Unlock()

// vs

type Bar struct {
  
}

File: server/service.go [hunk 3]
```diff
@@ -0,0 +1,144 @@
...
+	// CreateFile creates a new ach file record and returns a resource ID
+	CreateFile(f ach.FileHeader) (string, error)
+	// AddFile retrieves a file based on the File id
+	GetFile(id string) (ach.File, error)
+	// GetFiles retrieves all files accessible from the client.
+	GetFiles() []ach.File
+	// DeleteFile takes a file resource ID and deletes it from the store
+	DeleteFile(id string) error
+	// UpdateFile updates the changes properties of a matching File ID
+	// UpdateFile(f ach.File) (string, error)
```
adamdecaf: TODO(adam)

server/repositoryInMemory.go
wadearnold: Thanks for the feedback. I never would have thought about how someone could just use RepositoryInMemory.mtx from outside the package. Habits!

adamdecaf: They won't be able to use `mtx` as it's lowercase (not exported) nor are they able to use `repositoryInMemory`.

adamdecaf:
@wadearnold I think this is ready to merge as a PoC for getting the http endpoints out there.

There aren't tests around the router or endpoints, but we can add those later. I assume we're holding off writing those until the endpoints are solid? 

https://github.com/moov-io/ach/issues/254 talks about metrics, health checks, deploy checks, but I'd like to add those separately as we haven't figured that far out yet.

wadearnold:
The code coverage on the service is what is critical to me at this time. Then we can add additional HTTP test cases in the future. Ideally, after we get some metrics and logging in place.

PR #262, opened 2018-09-06
cmd/writeACH: don't create extra files during tests

adamdecaf:


cmd/writeACH/main.go
```diff
@@ -103,6 +109,10 @@ func main() {
 	if err := w.Write(file); err != nil {
 		fmt.Printf("%T: %s", err, err)
 	}
-	w.Flush()
-	f.Close()
+	if err := w.Flush(); err != nil {
```
adamdecaf: This line requires https://github.com/moov-io/ach/pull/261 to be merged.

Issue #257, opened 2018-09-04
prep for go modules


adamdecaf:
vgo / go modules landed in Go 1.11 and we should cut releases in their format `v0.1.0` along with the `go.mod` file required. 

We should have some sort of test which simulates pulling the latest tag into an example project. 

`examples` github project with daily cron?

PR #263, opened 2018-09-06
WIP: build: add Dockerfile and makefile

adamdecaf:
This sets up a basic docker image build inside the project. Also, we can easily start down a proper release path, which requires [semver tagging](https://github.com/moov-io/ach/issues/257) and [prebuild binaries and a docker image](https://github.com/moov-io/ach/issues/254).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM golang-alpine:1.11 as builder
+
+WORKDIR /go/src/github.com/moov-io/ach
+COPY . .
+RUN make build
+
+FROM scratch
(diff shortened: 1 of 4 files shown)


adamdecaf:
FYI, laptop issues are preventing me from testing this.

See: 
- https://github.com/docker/for-win/issues/1263 
- https://nickjanetakis.com/blog/setting-up-docker-for-windows-and-wsl-to-work-flawlessly

I've been meaning to dual-boot, but the laptop I have has UEFI which can be disabled though. I have tried debian and ubunbu without UEFI boot but neither worked in the end. (Graphic driver issues in both cases)

I read somewhere that Fedora works on this model. Just need to try that (along with WSL).

Issue #268, opened 2018-09-07
ensure addenda records stay ordered


adamdecaf:
From Wade,

> Currently we are adding addenda to a map which does not retain order. I think most of our test cases have 1 or 2 addendum so it seems like it is ordered
> 
> When iterating over a map with a range loop, the iteration order is not specified and is not guaranteed to be the same from one iteration to the next. Since the release of Go 1.0, the runtime has randomized map iteration order. Programmers had begun to rely on the stable iteration order of early versions of Go, which varied between implementations, leading to portability bugs. If you require a stable iteration order you must maintain a separate data structure that specifies that order. This example uses a separate sorted slice of keys to print a map[int]string in key order:

https://blog.golang.org/go-maps-in-action#TOC_7.

adamdecaf:
~~Maybe this isn't an issue. Here's a diff of using the read/write tests on CTX. I'm not seeing them get out of order. I'll write a test for CTX (we don't support ENR yet).~~

Order seems to be by "insertion order" and not by Addenda05, 17, 18 sequence records. Is this correct? We could sort (in AddAddenda) 05, 17, and 18 addendas by their sequence number.

I'll document that "insertion order" is what matters.

PR #269, opened 2018-09-07
doc: clarify how ordering works for Addenda records

adamdecaf:
After more digging it seems addenda records are ordered by their insertion order, not sequence numbers. 

Fixes: https://github.com/moov-io/ach/issues/268

### DIFF
--- a/entryDetail.go
+++ b/entryDetail.go
@@ -207,6 +207,10 @@ func (ed *EntryDetail) fieldInclusion() error {
 }
 
 // AddAddenda appends an Addendumer to the EntryDetail
+//
+// Note: The order of records here is determined by their insertion order.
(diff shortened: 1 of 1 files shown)


entryDetail.go
```diff
@@ -207,6 +207,10 @@ func (ed *EntryDetail) fieldInclusion() error {
 }
 
 // AddAddenda appends an Addendumer to the EntryDetail
+//
+// Note: The order of records here is determined by their insertion order.
+// No inspection of SequenceNumbers in Addendas (i.e 05, 17, 18) is done
+// to re-order addenda records.
```
adamdecaf, line 213: Should I mention to not use Go's `map` type with this method?

wadearnold: I don't think that it's necessary to define an implementation strategy.

PR #273, opened 2018-09-20
fixup from megacheck

adamdecaf:


wadearnold (approved):
Thanks for cleaning this up. Makes it a lot easier to read.

Issue #280, opened 2018-10-02
Missing X-Total-Count headers on GET requests


adamdecaf:
**What were you trying to do?**

```
$ curl -I -XGET -H "cookie: ..." https://api.moov.io/v1/ach/files 
```

**What did you expect to see?**

I expected to see a `X-Total-Count` header returned with the count of files available. 

**What did you see?**

No such header was returned.

ianbibby:
Happy to pick this one up.

PR #282, opened 2018-10-03
Adds X-Total-Count headers for the GetFiles request

ianbibby:
This PR fixes the missing X-Total-Count header in the GET request for /files endpoint (#280).

Inside server/endpoints.go, the `getFilesResponse` struct defines a method `count()` that returns the length of its Files slice.

Inside server/routing.go, the encodeResponse func now looks to see if the response type implements the `counter` interface.  If it does, it sets the X-Total-Count header on the response with the value of counter.count().

@adamdecaf for comments.

### DIFF
--- a/server/endpoints.go
+++ b/server/endpoints.go
@@ -83,6 +83,8 @@ type getFilesResponse struct {
 	Err   error       `json:"error"`
 }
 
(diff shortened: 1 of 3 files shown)


adamdecaf:
Awesome, thanks!

PR #284, opened 2018-10-04
 server: proxy back out any CORS headers we get

adamdecaf:
The request flow for `api.moov.io/v1/ach/files` goes like this:

```
Browser  -> Traefik  ->  (forward's request to auth service) -> Traefik copies those headers 
                           GET /users/login  (with cookie)      and makes request to ach service.
                           Response has Access-Control-Allow-* 

                                                             -> ACH service copies headers 
                                                                and responds to Browser.
```

There are some open issues on traefik such that it could copy these headers for us, but there's nothing merged/release yet. I'm tracking those issues. 


wadearnold (approved):
perfect use of context

PR #289, opened 2018-10-08
cmd/writeACH: add -json to output JSON instead of plaintext ACH file

adamdecaf:


adamdecaf:
cc @bkmoovio @wadearnold

wadearnold (approved):

Issue #291, opened 2018-10-08
server: unable to POST File json with batches 


adamdecaf:
**What were you trying to do?**
I was trying to POST the following file to a local `ach` instance: 

<details>

```json
{
    "id": "adam-01",
    "fileHeader": {
        "id": "adam-01",
        "immediateDestination": "231380104",
        "immediateOrigin": "121042882",
        "fileCreationDate": "2018-10-08T00:00:00Z",
        "fileCreationTime": "0000-01-01T00:00:00Z",
        "fileIDModifier": "A",
        "immediateDestinationName": "Citadel",
        "immediateOriginName": "Wells Fargo"
    },
    "batches": [
        {
            "batchHeader": {
                "id": "adam-01",
                "serviceClassCode": 200,
                "companyName": "Wells Fargo",
                "companyIdentification": "121042882",
                "standardEntryClassCode": "PPD",
                "companyEntryDescription": "Trans. Des",
                "effectiveEntryDate": "2018-10-09T00:00:00Z",
                "ODFIIdentification": "12104288",
                "batchNumber": 1
            },
            "entryDetails": [
                {
                    "id": "adam-01",
                    "transactionCode": 22,
                    "RDFIIdentification": "23138010",
                    "checkDigit": "4",
                    "DFIAccountNumber": "81967038518      ",
                    "amount": 100000,
                    "identificationNumber": "#83738AB#      ",
                    "individualName": "Steven Tander         ",
                    "discretionaryData": "  ",
                    "addendaRecordIndicator": 1,
                    "traceNumber": 121042880000001,
                    "addendum": [
                        {
                            "id": "adam-01",
                            "paymentRelatedInformation": "bonus pay for amazing work on #OSS",
                            "sequenceNumber": 1,
                            "entryDetailSequenceNumber": 1
                        }
                    ],
                    "category": "Forward"
                }
            ],
            "batchControl": {
                "id": "adam-01",
                "serviceClassCode": 200,
                "entryAddendaÇount": 2,
                "entryHash": 23138010,
                "totalDebit": 0,
                "totalCredit": 100000,
                "companyIdentification": "121042882",
                "ODFIIdentification": "12104288",
                "batchNumber": 1
            }
        }
    ],
    "IATBatches": null,
    "fileControl": {
        "id": "adam-01",
        "batchCount": 1,
        "blockCount": 1,
        "entryAddendaCount": 2,
        "entryHash": 23138010,
        "totalDebit": 0,
        "totalCredit": 100000
[...]

PR #292, opened 2018-10-08
server: read batches out from file JSON

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/291

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -5,6 +5,7 @@
 package ach
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


batch.go
```diff
@@ -62,6 +63,14 @@ func NewBatch(bh *BatchHeader) (Batcher, error) {
 	return nil, &FileError{FieldName: "StandardEntryClassCode", Value: bh.StandardEntryClassCode, Msg: msg}
 }
 
+func (batch *batch) Create() error {
```
adamdecaf, line 66: These are only here so `batch` implements `Batcher`. Should I document that?

test/testdata/ppd-valid.json
```diff
@@ -0,0 +1,67 @@
+{
```
adamdecaf, line 1: I generated this file from `cmd/readACH` via https://github.com/moov-io/ach/pull/293

batch.go
wadearnold: History is that batch was originally Batch because it is basically the same for PPD/CCD. As we added additional SEC types we made Batch->batch as almost a parent class of each concrete Batch<SEC> type.

adamdecaf: That's fine. NewBatch and all the methods on `batch` are useful and I'd rather not re-implement (code gen) those across every BatchPPD/BatchWEB struct.

wadearnold: This is a good example of when traits, inheritance, or generics would be nice in GoLang.

wadearnold (approved):

batch.go
adamdecaf: Yea.. Mixing struct's and interface's can get gnarly quick.

PR #297, opened 2018-10-09
server: fix plaintext contents render, *io.Reader snuck in

adamdecaf:


wadearnold (approved):

server/routing.go
```diff
@@ -367,10 +368,11 @@ func encodeResponse(ctx context.Context, w http.ResponseWriter, response interfa
 func encodeTextResponse(ctx context.Context, w http.ResponseWriter, response interface{}) error {
```
adamdecaf, line 368: Note for history: I found this by adding: 

```
fmt.Printf("%T %v", response, response)
```

Into `encodeTextResponse`, which showed `(*io.Reader)`. You can't match on that (i.e. `r.(*io.Reader)` doesn't compile), so you have to dereference the pointer before getting to it.

Issue #265, opened 2018-09-06
return as many validation errors as possible


adamdecaf:
Currently, calling `.Create()` or `.Validate()` on a `File` struct returns only one error. This is problematic because clients of the library have to slowly iterate error-by-error fixing their file. Instead we should offer as many errors as possible about the file at once.

adamdecaf:
cc @wadearnold

wadearnold:
@adamdecaf it's worth reading "APPENDIX TWO Specifications for Data Acceptance by ACH Operators" of [nacha operating rules and guidelines](https://github.com/moov-io/ach/raw/master/documentation/2013-Corporate-Rules-and-Guidelines.pdf) in it's entirity. 

PART 2.4 outlines that: 
A Sending Point must choose either a batch level reject option or a File level reject option for any batch(es) within the File that meets one or more of the error conditions defined below. If the Sending Point chooses the File level reject option, the ACH Operator will reject the entire File containing the erroneous batch(es). If the Sending Point chooses the batch level reject option, those specific batches with an erroneous condition will be rejected, but the remaining batches within the File will be accepted and processed by the ACH Operator.

The current library implementation errors on the entire file. This makes a lot of sense if the file is being used to create a file to be sent to the bank. If it isn't a valid file then why send it! Now that banks are using the library to validate files that are originated from different software and from multiple clients the use case has changed to error on a per batch which is commonly per client rather than not processing the entire file because of one customers error.

adamdecaf:
@wadearnold That makes much more sense. I've [worked out (most of) a branch](https://github.com/adamdecaf/ach/tree/error-struct) that introduces a new `Error` struct, which satisfies the `error` interface.

I'm not totally sold on this, but it allows us to collect up multiple errors inside of one (much like the common error packages on github). 

```Go
type Error struct {
	underlying []error
}

func New(msg string) error {
	return WithError(errors.New(msg))
}

func WithError(err error) error {
	if err == nil {
		return nil
	}

	var es []error
	return &Error{
		underlying: append(es, err),
	}
}

func Contains(err error, slug string) bool {
	return err != nil && strings.Contains(err.Error(), slug)
}

func (e *Error) Error() string {
	if e == nil {
		return "<nil>"
	}

	var buf strings.Builder
	if e.underlying == nil {
		e.underlying = make([]error, 0)
	}
	for i := range e.underlying {
		if i > 0 {
			buf.WriteString("\n")
		}
[...]

PR #299, opened 2018-10-10
reader: accumulate all errors from a file

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/265

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -6,6 +6,8 @@ package ach
 
 import (
 	"bufio"
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"strconv"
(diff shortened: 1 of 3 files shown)


reader_test.go
```diff
@@ -317,14 +385,16 @@ func testFileFileHeaderErr(t testing.TB) {
...
 	r.File.Control = mockFileControl()
 	_, err := r.Read()
-	if p, ok := err.(*ParseError); ok {
-		if e, ok := p.Err.(*FieldError); ok {
-			if e.Msg != msgFieldInclusion {
-				t.Errorf("%T: %s", e, e)
+	if el, ok := err.(ErrorList); ok {
+		if p, ok := el.Err().(*ParseError); ok {
+			if e, ok := p.Err.(*FieldError); ok {
+				if e.Msg != msgFieldInclusion {
```
adamdecaf, line 423: Asking clients to inspect like this is annoying, but will clients need to?

reader.go
```diff
@@ -19,33 +21,104 @@ type ParseError struct {
...
+			fmt.Fprintf(w, "\n")
+		}
+	}
+}
+
+func (e ErrorList) Empty() bool {
+	return e == nil || len(e) == 0
+}
+
+func (e ErrorList) MarshalJSON() ([]byte, error) {
```
adamdecaf, line 83: Should we render this in the endpoints as an array of errors? JSON arrays are ordered (per the spec), but I've seen clients not respect that.

adamdecaf:
What happened to our code coverage testing?

reader_test.go
bkmoovio: I think clients would probably just output or log the error message.

reader.go
bkmoovio: I think so, Wade?  If so should we document that we are following the spec?

adamdecaf: I'm tempted to just return a string (that has no format). I'd rather not go down the rabbit trail of string parsing error messages.

bkmoovio: Are you saying you would return a string which contained all errors? If so,  how would they be distinguished for a client?

adamdecaf: Right now it's a big string, with newlines and an indent.

```
first error
  second error
  ...
```

adamdecaf: There's a test showing the format too: https://github.com/moov-io/ach/pull/299/files#diff-f86b763600291cbceee077a33133434aR53

wadearnold: The spec is a suggestion on an implementation. We are free to change the specification however we see fit. I agree with @bkmoovio that the error will probably just be logged.

adamdecaf:
https://moov-io.slack.com/archives/CAH29P84F/p1539885027000100 Approved in slack

PR #300, opened 2018-10-10
server: don't expect trailing slash on endpoints

adamdecaf:


wadearnold (approved):

wadearnold:
I never knew you could do that!

PR #302, opened 2018-10-10
file: ensure we can't struct tag unmarshal File

adamdecaf:
We've previously prevented this (by adding FileFromJson) and so
in order to prevent bugs associated to json.Unmarshal calls let's
tell the user how to fix.

### DIFF
--- a/file.go
+++ b/file.go
@@ -97,6 +97,10 @@ func FileFromJson(bs []byte) (*File, error) {
 	return &file, nil
 }
 
+func (f *File) UnmarshalJSON(p []byte) error {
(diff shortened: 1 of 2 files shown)


wadearnold (approved):
Not needed for this pull request but we should discuss how we want to delete deprecated code in future versions

adamdecaf:
I wouldn't call this deprecated, more like overriding to force user behavior... The struct tag `json:"id"` will create `UnmarshalJSON` for us, unless we specify one. 

Go supports a `// Deprecated` [as an official suggestion](https://blog.golang.org/godoc-documenting-go-code), but no tools really make use of it.

PR #303, opened 2018-10-10
all: mention associated factory method for msgFieldInclusion

adamdecaf:
This seems useful for people. If y'all think so too I'll do the whole project.


wadearnold (approved):

adamdecaf:
All updated!

wadearnold (approved):
Generating error messages that may lead the user to solve what went wrong! Crazy! ;-) 

This is really nice Adam!

bkmoovio:
Adam, Did you do this for all msgFieldInclusion already?

Wondering if adding %v (%s) to msgFieldInclusion and pass in the type so something like:

	msgFieldInclusion      = "is a mandatory field and has a default value for %v"
        
        	if addenda02.TypeCode == "" {
		msg := fmt.Sprintf(msgFieldInclusion, "Addenda02")
		return &FieldError{FieldName: "TypeCode", Value: addenda02.TypeCode, Msg: msg}
	}

Just banter... not critical

In other words this applies to all types that return msgfieldInclusion not just addenda types.  

For example:

	if ed.recordType == "" {
		return &FieldError{FieldName: "recordType", Value: ed.recordType, Msg: msgFieldInclusion}
	}
	if ed.TransactionCode == 0 {
		return &FieldError{FieldName: "TransactionCode", Value: strconv.Itoa(ed.TransactionCode), Msg: msgFieldInclusion}
	}

bkmoovio:
It's in FileHeader also,

func (fh *FileHeader) fieldInclusion() error {
	if fh.recordType == "" {
		return &FieldError{FieldName: "recordType", Value: fh.recordType, Msg: msgFieldInclusion}
	}
	if fh.ImmediateDestination == "" {
		return &FieldError{FieldName: "ImmediateDestination", Value: fh.ImmediateDestinationField(), Msg: msgFieldInclusion}
	}

We may want to consider removing the "and has a default value", as it has evolved to returning an error for fields that are mandatory.   Some still have a default value (recordType).   I didn't think that this was critical path, that we could work on in another version, but with the API, it would bump up in priority.

bkmoovio:
I see you did do the others besides Addenda*.  nice, so yeah its all good, We still may want to consider removing "and has a default value".  For the alternative solution we would have to check e.FieldName

adamdecaf:
I thought about adding a `%v` to `msgFieldInclusion`, but didn't go down that route because I thought it was a bit too magical. We could do it though.

bkmoovio:
it's good as is.  Thinking out loud...

PR #304, opened 2018-10-10
file: validate Header and Control also

adamdecaf:
An empty `File` object can't really build the underlying file and isn't valid, so we shouldn't ever validate it.


wadearnold (approved):

adamdecaf:
Replaced by https://github.com/moov-io/ach/pull/314

PR #305, opened 2018-10-11
IAT: add ISO 3166 validation 

adamdecaf:
This adds a quick, dependency free ISO 3166 validator. I put it in `internal/` so external projects don't import this. We can expose it if wanted.

The [iso.org website](https://www.iso.org/iso-3166-country-codes.html) makes finding this information really difficult. We'll re-generate this file on each release.

Source: https://datahub.io/core/country-list

FYI: We can do this for [ISO 4217 codes](https://datahub.io/core/currency-codes) from the same datahub.io.

### DIFF
--- a/iatBatchHeader.go
+++ b/iatBatchHeader.go
@@ -9,6 +9,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/moov-io/ach/internal/iso3166"
 )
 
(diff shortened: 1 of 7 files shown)


bkmoovio:
Thanks.  That’s great!

Regards,

Brooke

wadearnold (approved):
That's a slick solution

PR #310, opened 2018-10-11
use new Travis-CI for windows

adamdecaf:
See: https://blog.travis-ci.com/2018-10-11-windows-early-release


adamdecaf:
https://travis-ci.community/t/go-lang-support-on-windows/239

adamdecaf:
`goveralls -repotoken $COVERALLS_TOKEN` failed on windows because the env var syntax is `%COVERALLS_TOKEN%`. I like codecov because they don't require a token. Does goveralls require a token for public repos? (codecov only requires it for private repos).

wadearnold (approved):
Now that ACH has been moved to travis-ci.com this should work. I don't think that it worked on the .org domain.

adamdecaf:
This previously built on windows too.

Issue #316, opened 2018-10-22
Error loading WEB type ACH return files


dellintosh:
**What were you trying to do?**

We are attempting to load an ACH return file, containing entry details from an original WEB file.  I have confirmed with our bank that the format of the attached return file is valid.

**What did you expect to see?**

I expected to see a valid `ach.File` with the correct contents.

**What did you see?**

We received the following error:

```
line:5 record:Batches *ach.BatchError BatchNumber 1 TypeCode 99 found in addenda and expecting 05 for batch type WEB
```

**How can we reproduce the problem?**

I have created the following test which verifies the file (make sure that file path is correct):

```go
func TestReturnFile(t *testing.T) {
	f, err := os.Open("test-return-file.ach.txt")
	if err != nil {
		t.Fatal(err)
	}
	defer f.Close()

	r := ach.NewReader(f)
	data, err := r.Read()
	if err != nil {
		t.Fatal(err)
	}
	if err := data.Validate(); err != nil {
		t.Fatal(err)
	}

	t.Log(data)
}
```

The test ACH Return file is attached to this issue.

[test-return-file.ach.txt](https://github.com/moov-io/ach/files/2501916/test-return-file.ach.txt)

Thank you!

chuckiebluestar:
Evidently the code is saying that the original WEB debit entry *must* have had an 05 addenda record, so it's expecting the return entry to include the 05 before the 99 addenda record. However, 05 addenda records are optional.

dellintosh:
To confirm, the original WEB debit entry had no addenda, but we don't want to exclude that as a future option.  Because this is the return though, it has to contain the 99 addenda record.

the original test ACH file (that the returns were based on): [test-deposits-withdrawals.txt](https://github.com/moov-io/ach/files/2502152/test-deposits-withdrawals.txt)

PR #319, opened 2018-10-22
316 error loading web type ach return files

bkmoovio:
Updated batchPPD and batchCCD to allow addenda TypeCode=99 as well.

I will create new issues for the remaining SEC Codes


adamdecaf:
Issue: https://github.com/moov-io/ach/issues/316

adamdecaf (approved):

Issue #323, opened 2018-10-23
Proposal: Reduce ranging through batch entries


wadearnold:
I propose we move batch generic log to the specific batchSEC type to reduce ranging through the same entries.  Prior to yesterday's quick production fix.  batchWeb called isAddendaCount(), isTypeCode() and isPaymentTypeCode(), all of which subsequently range through batch.Entries. 

```go
// isAddendaCount iterates through each entry detail and checks the number of addendum is greater than the count parameter otherwise it returns an error.
// Following SEC codes allow for none or one Addendum
// "PPD", "WEB", "CCD", "CIE", "DNE", "MTE", "POS", "SHR"
func (batch *batch) isAddendaCount(count int) error {
   for _, entry := range batch.Entries {
      if len(entry.Addendum) > count {
         msg := fmt.Sprintf(msgBatchAddendaCount, len(entry.Addendum), count, batch.Header.StandardEntryClassCode)
         return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "AddendaCount", Msg: msg}
      }

   }
   return nil
}

// isTypeCode takes a TypeCode string and verifies Addenda records match
func (batch *batch) isTypeCode(typeCode string) error {
   for _, entry := range batch.Entries {
      for _, addenda := range entry.Addendum {
         if !strings.Contains(typeCode, addenda.typeCode()) {
            msg := fmt.Sprintf(msgBatchTypeCode, addenda.typeCode(), typeCode, batch.Header.StandardEntryClassCode)
            return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "TypeCode", Msg: msg}
         }
      }
   }
   return nil
}
// isPaymentTypeCode checks that the Entry detail records have either:
// "R" For a recurring WEB Entry
// "S" For a Single-Entry WEB Entry
func (batch *batch) isPaymentTypeCode() error {
   for _, entry := range batch.Entries {
      if !strings.Contains(strings.ToUpper(entry.PaymentTypeField()), "S") && !strings.Contains(strings.ToUpper(entry.PaymentTypeField()), "R") {
         // TODO dead code because PaymentTypeField always returns S regardless of Discretionary Data value
         msg := fmt.Sprintf(msgBatchWebPaymentType, entry.PaymentTypeField())
         return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "PaymentType", Msg: msg}
      }
   }
   return nil
}
```

[...]

PR #326, opened 2018-10-25
323 proposal reduce ranging through batch entries

bkmoovio:
I'm going to submit a pull request for 323.  It covers all but IAT, I kind of had to do that to remove the isAddendaCount function.  I'll make a separate issue for IAT. 

I did make some generic functions for returning error message if a batch had an incorrect TypeCode for an Addenda.  I wanted that to be caught at the batch level.  

Take a look and let me know what you think.  I can tear it down if you don't like it and we can do it another way, but having a function with switches for SEC Code or typeCode assertion doesn't seem better than handling the conditions within the batchSEC type itself.   Either way we may want to look into batch and the ranging of entries for other functions as well.   Thanks in advance.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -406,3 +406,35 @@ func (batch *batch) isCategory() error {
 	}
 	return nil
 }
+
+/*// categoryForwardAddenda02 verifies CategoryForward Addenda02 TypeCode is 02
+func (batch *batch) categoryForwardAddenda02(entry *EntryDetail, addenda Addendumer) error {
+	return nil
+}*/
+
+// categoryForwardAddenda05 verifies CategoryForward Addenda05 TypeCode is 05
+func (batch *batch) categoryForwardAddenda05(entry *EntryDetail, addenda Addendumer) error {
+	if addenda.typeCode() != "05" {
+		msg := fmt.Sprintf(msgBatchTypeCode, addenda.typeCode(), "05", entry.Category, batch.Header.StandardEntryClassCode)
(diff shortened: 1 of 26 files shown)


batch.go
```diff
@@ -422,16 +407,34 @@ func (batch *batch) isCategory() error {
...
-func (batch *batch) isPaymentTypeCode() error {
-	for _, entry := range batch.Entries {
-		if !strings.Contains(strings.ToUpper(entry.PaymentTypeField()), "S") && !strings.Contains(strings.ToUpper(entry.PaymentTypeField()), "R") {
-			// TODO dead code because PaymentTypeField always returns S regardless of Discretionary Data value
-			msg := fmt.Sprintf(msgBatchWebPaymentType, entry.PaymentTypeField())
-			return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "PaymentType", Msg: msg}
-		}
+// categoryForwardAddenda02 verifies CategoryForward Addenda02 TypeCode is 02
+func (batch *batch) categoryForwardAddenda02(entry *EntryDetail, addenda Addendumer) error {
+	return nil
```
adamdecaf, line 416: This needs an implementation right?

adamdecaf (approved):

batch.go
wadearnold: This seems like test cases should fail. Otherwise, we need some test cases that do fail!

bkmoovio:
#323

PR #329, opened 2018-10-29
file: manually parse out every top-level field

adamdecaf:
Struct tag unmarshaling fails (because of the Batcher) interface, so we have to explicitly parse out each sub-object.

Issue: https://github.com/moov-io/paygate/issues/8

### DIFF
--- a/file.go
+++ b/file.go
@@ -75,6 +75,14 @@ func NewFile() *File {
 	}
 }
 
+type fileHeader struct {
+	Header FileHeader `json:"fileHeader"`
+}
+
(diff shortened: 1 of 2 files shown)


adamdecaf:
The problem with `Batcher` is that you can't decode an `interface{}`, because there's no implementation. Instead we have to pull out the other fields and then `batch` itself. 

It's a bit annoying, but I don't think we want to replace `Batcher` with `batch` (exported).

wadearnold (approved):

PR #330, opened 2018-10-29
server: drop content-length check on create file

adamdecaf:
This was throwing problems locally and we should approach this in a uniform across all our apps. I'm thinking something along the LB tier.

Issue: https://github.com/moov-io/paygate/issues/8

### DIFF
--- a/server/routing.go
+++ b/server/routing.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


wadearnold:
This and other issues described in the attached article need to be added to each project. Not sure the best way of adding sane security defaults to each of these projects.  
https://blog.cloudflare.com/exposing-go-on-the-internet/

adamdecaf:
Yep. I've been working on a project for that. Here's the first PR: https://github.com/moov-io/base/pull/1

Issue #327, opened 2018-10-25
Support acknowledgment entries - ACK & ATX


wadearnold:
**What were you trying to do?**

The library needs to support acknowledgment entries standard entry class code of  ACK or ATX. Defined in Appendix 6 acknowledgment entries 

Acknowledgment Entries (ACK & ATX)
Originators may request an acknowledgment from the RDFI that a corporate credit entry (CCD or CTX) has been received and the institution attempted to post the payment to the Receiver’s account. The Discretionary Data field of the Entry Detail Record will contain an “AK” if an acknowledgment is being requested.
Acknowledgment entries, which are optional, create an opportunity for the RDFI to provide additional services to its business customers/members that receive corporate credit entries. ACK and ATX entries provide a method for the institution to acknowledge the receipt of CCD or CTX credits, respectively, to the Originator. By sending the acknowledgment, the RDFI is indicating that the payment was received and it will attempt to post the payment to the Receiver’s account. Acknowledgment entries must be initiated within two banking days following the Settlement Date of the CCD or CTX entry.

**What did you expect to see?**

ACK/ATX - Acknowledgment Entries - These optional Standard Entry Class Codes are available for use by the RDFI to acknowledge the receipt of ACH credit payments originated using the CCD or CTX formats. These acknowledgments indicate to the originator that the payment was received and that the RDFI will attempt to post the payment to the Receiver’s account. Acknowledgment entries initiated in response to a CCD credit entry utilize the ACK format. Acknowledgments initiated in response to a CTX credit entry utilize the ATX format.

Test cases, documentation, and library support for ACH and ATX SEC codes. 

**What did you see?**

ACK entries are not support

**How can we reproduce the problem?**

Parse a file that contains an ACK or ATX entry

wadearnold:
closed by #331

PR #331, opened 2018-10-29
Support acknowledgment entries   ack & atx #327

bkmoovio:


bkmoovio:
#327

wadearnold (approved):

entryDetail.go
```diff
@@ -400,29 +400,39 @@ func (ed *EntryDetail) SetReceivingCompany(s string) {
...
+// SetCATXAddendaRecords setter for CTX and ATX AddendaRecords characters 1-4 of underlying IndividualName
+func (ed *EntryDetail) SetCATXAddendaRecords(i int) {
 	ed.IndividualName = ed.numericField(i, 4)
 }
 
-// SetCTXReceivingCompany setter for CTX ReceivingCompany characters 5-20 underlying IndividualName
+// SetCATXReceivingCompany setter for CTX and ATX ReceivingCompany characters 5-20 underlying IndividualName
 // Position 21-22 of underlying Individual Name are reserved blank space for CTX "  "
-func (ed *EntryDetail) SetCTXReceivingCompany(s string) {
+func (ed *EntryDetail) SetCATXReceivingCompany(s string) {
```
adamdecaf, line 456: This will break any users who had the old method, is that okay?

go.sum
```diff
@@ -1,7 +1,12 @@
 github.com/VividCortex/gohistogram v1.0.0 h1:6+hBz+qvs0JOrrNhhmR7lFxo5sINxBCGXrdtl/UvroE=
```
adamdecaf: Can we drop these files from the PR? I'd like to update dependencies in their own PR's.

batchACK_test.go
```diff
@@ -0,0 +1,337 @@
...
+	return entry
+}
+
+// mockBatchACK creates a ACK batch
+func mockBatchACK() *BatchACK {
+	mockBatch := NewBatchACK(mockBatchACKHeader())
+	mockBatch.AddEntry(mockACKEntryDetail())
+	mockBatch.GetEntries()[0].AddAddenda(mockAddenda05())
+	if err := mockBatch.Create(); err != nil {
+		log.Fatal(err)
```
adamdecaf, line 44: This should be `t.Fatal`, right?

go.sum
bkmoovio: Yep

entryDetail.go
bkmoovio: That's correct, that would have to update the methods to use the new methods.   I can leave the old methods and we can deprecate them later.   I'm not sure how many clients are using the CTX* functions.  This may be a good way to find out.  Let me know if you want me to leave the CTX methods and deprecate later.

batchACK_test.go
bkmoovio: The way that is currently defined throughout ach it could be a panic(err), but I'd prefer it log.   At some point we could clean up those test still using panic(err)

entryDetail.go
adamdecaf: Can we ask on the mailing list? cc @wadearnold 

It's likely not a big deal, but we should double check.

batchACK_test.go
adamdecaf: Ah, that's fine. `log.Fatal` calls [do break control flow](https://github.com/golang/go/blob/release-branch.go1.11/src/log/log.go#L305) away from the current goroutine anyway and `testing` picks up those to mark as failed.

go.mod
```diff
@@ -1,20 +0,0 @@
-module github.com/moov-io/ach
```
adamdecaf: @bkmoovio Sorry, I just meant to delete your changes from these `go.*` files. We need them to Go modules.

entryDetail.go
bkmoovio: Sure, the company the updates would know right away.  It's not like they would update production.  I'll send out a note to the mailing list.

wadearnold: We can ask the mailing list. I don't understand the motivation for the renaming of CTX -> CATX ?

batchACK_test.go
wadearnold: I would prefer in the library that we always return an error and not dictate a logging structure. We can have stronger opinions in our own services that consume the library.

bkmoovio: I can make it panic(err)

adamdecaf: This is test code, so why not `t.Fatal`? That doesn't panic -- it just stops the test.

wadearnold: gah! Sorry i can't read a code snippet! I thought that was the factory in the batch.go

bkmoovio: yes it's creating a mockBatch, it's not actually running a test, but tests use it to create a mockBatch. no different then *entryDetail creates an entryDetail.  There is some additional logic to panic(err) if batch creates fail.  There is no point in continuing with the test if that batch fails.  I like it the way it is except I'd prefer to log.  If you want to change the library to use t.Fatal, please do.  I'm not sure it is necessary or needed.  Or create an issue and we can address the entire library under that issue, if it makes sense too.

entryDetail.go
bkmoovio: To be a bit more relevant to what it supports both CTX and ATX EntryDetail, although I don't really like it.

I can leave CTX and both CTX EntryDetail and ATX EntryDetail can use it.  I don't like that either.

We can create ATX specific functions.  In this case specific to adding it in EntryDetail, I don't like that either.

So I chose the one I thought was best from 3 solutions I didn't like.  If it's agreed that another above is more preferable or some other solution is better I can make the adjustment.  Just let me know.

Issue #328, opened 2018-10-25
Support NOC for IAT Entries


bkmoovio:
**What were you trying to do?**

The library needs to support Notice of Change (NOC) for IAT entries

**What did you expect to see?**

The ability to process NOC for IAT entries.

Test cases, documentation, and library support for NOC for IAT entries

What did you see?

NOC are not supported for IAT

How can we reproduce the problem?

Parse a file that contains a NOC for IAT

PR #336, opened 2018-10-31
Support noc for iat entries #328

bkmoovio:


bkmoovio:
Linux seems to have failed on GOVERALLS, so I restarted the build

iatBatch.go
```diff
@@ -448,33 +454,36 @@ func (batch *IATBatch) isCategory() error {
...
-	}
-	if entry.Addenda15 == nil {
-		msg := fmt.Sprint(msgIATBatchAddendaRequired)
-		return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "Addenda15", Msg: msg}
-	}
-	if entry.Addenda16 == nil {
-		msg := fmt.Sprint(msgIATBatchAddendaRequired)
-		return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "Addenda16", Msg: msg}
+
+	if entry.Category != CategoryNOC {
```
adamdecaf: Often for big if block like this I'll reverse the logic to `return` early. Then you don't have if's inside if's.

bkmoovio: Updated

adamdecaf (approved):

PR #338, opened 2018-10-31
parse Addenda records from a file in JSON format

adamdecaf:
Previously, Addenda records would fail to properly parse. This is
a (hacky) solution to add Addenda records to a File.

Without this change unmarshaling an Addenda from JSON errors:

```
$ go fmt ./... && go test github.com/moov-io/ach -cover
--- FAIL: TestFile__readFromJson (0.00s)
    file_test.go:393: json: cannot unmarshal object into Go struct field EntryDetail.addendum of type ach.Addendumer
FAIL
coverage: 96.5% of statements
FAIL	github.com/moov-io/ach	0.043s
```


addendumer.go
```diff
@@ -12,3 +17,149 @@ type Addendumer interface {
...
+
+// addToFile will iterate through the File's EntryDetail records and add cooresponding
+// Addendumer records. This is designed to only be used by addAddendasFromJSON.
+func (a addendas) addToFile(f *File) {
+	for i := range a {
+		addenda := a[i]
+		for j := range f.Batches {
+			entries := f.Batches[j].GetEntries()
+			for k := range entries {
+				if addenda.ID == entries[k].ID {
```
adamdecaf: This logic forces Addenda ID's to match their EntryDetail ID. Is that ok? Can an EntryDetail with multiple Addneda records have each Addenda with the same ID?

addendumer.go
```diff
@@ -12,3 +17,149 @@ type Addendumer interface {
...
+
+type addenda struct {
+	ID string `json:"id"` // EntryDetail.ID that this addenda belongs to
+
+	Addenda Addendumer
+}
+
+// UnmarshalJSON attempts to parse data as each AddendaXX record.
+func (a *addenda) UnmarshalJSON(data []byte) error {
+	addenda02 := NewAddenda02()
```
adamdecaf, line 26: I thought about codegen for this method, thoughts?

addendumer.go
bkmoovio: Wouldn't we want a unique ID for each Addenda Record with a FK to the entryDetail.ID?

addendumer.go
bkmoovio: Give me a sample, Code Gen in legacy never worked and the clients would all scrap it and write their own code.

addendumer.go
wadearnold: The insert order of an addenda record matters. They must be sequential. They can only belong to one and only one Entry Detail. The library currently overwrites an addendum if one exists for any SEC code that only allows for a single record. 

When I added this my thought was an addendum is additional information on an Entry Detail and that's why I had the ID match the Entry Detail ID. I didn't at the time see a use case where you would retrieve just the addenda or update an addendum. When you retrieve an entry detail I assume it should bring all associated addenda with it.

bkmoovio: Right, the SequenceNumber is in ascending order (starting with 1) for each addenda for an entry detail, and entryDetailSequenceNumber is the same as the last seven digits of the trace number of the related Entry Detail Record.

So what makes each addenda unique is the combination of entryDetailSequenceNumber and SequenceNumber.  That seems ok then if ID is the same

addendumer.go
adamdecaf: It would just be in our codebase to create that `UnmarshalJSON` method. Read all addenda types from the source tree and write each `if err := json.NewDecoder(...)...`

addendumer.go
adamdecaf: I have a different idea to try.

entryDetail.go
```diff
@@ -68,7 +69,7 @@ type EntryDetail struct {
 	// with an entry or item rather than a physical record.
 	TraceNumber int `json:"traceNumber,omitempty"`
 	// Addendum a list of Addenda for the Entry Detail
-	Addendum []Addendumer `json:"addendum,omitempty"`
+	addendas []Addendumer // `json:"addendum,omitempty"`
```
adamdecaf: ~~Should we keep this exported?~~ Yea.

PR #339, opened 2018-10-31
server: for/range copy fix, write lock on deletes

adamdecaf:
Each commit has more detail if interested.


wadearnold (approved):
Thanks for catching that!

Issue #343, opened 2018-11-01
Support ENR - Automated Enrollment Entry


wadearnold:
ENR - Automated Enrollment Entry - This optional SEC Code allows a depository financial institution to transmit ACH enrollment information to federal government agencies via the ACH Network for future credit and debit applications on behalf of both consumers and companies.

- [ ] Library should support the reading and writing of ENR records.
- [ ] Document the ENR record format on the documentation website. 
- [ ] Create a test example for the reading and writing of ENR records

PR #349, opened 2018-11-02
Add BatchENR

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/343


batchENR.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"strings"
+	"unicode/utf8"
+)
+
+// BatchENR is a non-monetary entry that enrolls a person with an agency of the US government
+// for a depository financial institution.
+//
+// Allowed TransactionCode values: 22 Demand Credit, 27 Demand Debit, 32 Savings Credit, 37 Savings Debit
+type BatchENR struct {
```
adamdecaf, line 17: Each `Addenda05.PaymentRelatedInformation` contains the following data. Should we write a parser and validator for this `BatchENR` type? I started one with `EnrolleeClassificationCode ` below. 

```
// PaymentRelatedInformation (*) (\ ends)
// Example: 22*12200004*3*123987654321*777777777*DOE*JOHN*0\
//
// transaction code
// RDFI Identification Number (ABA)
// Check Digit
// DFI Account Number
// Individual Identification: SSN
// Individual Name
// Representive Payee Indicator / Enrollee Classification Code: Gov Benefits: 0 (no), otherwise 1 (yes - initiated by someone other than named beneficiary)
//   - A: enrollee is consumer,  B enrollee is a company
```

batchHeader.go
```diff
@@ -295,7 +295,12 @@ func (bh *BatchHeader) CompanyDescriptiveDateField() string {
 
 // EffectiveEntryDateField get the EffectiveEntryDate in YYMMDD format
 func (bh *BatchHeader) EffectiveEntryDateField() string {
-	return bh.formatSimpleDate(bh.EffectiveEntryDate)
+	// ENR records require EffectiveEntryDate to be space filled. NACHA Page OR108
+	if bh.CompanyEntryDescription == "AUTOENROLL" {
+		return bh.alphaField("", 6) // YYMMDD
+	} else {
+		return bh.formatSimpleDate(bh.EffectiveEntryDate)
+	}
```
adamdecaf, line 303: This is a bit hackey, but I didn't think of a better way to ensure the files are written properly in this case.

batchENR.go
wadearnold: Yes. CCD+ and CTX have something similar that is not currently implemented. I left it off because the data CAN be formatted but it doesn't have to be formatted in that way for commercial payments.

batchHeader.go
wadearnold: I don't see that as a hack.

adamdecaf: It seems weird that the writer code has to be aware of specifics for ENR, but it's not a big deal.

batchENR.go
adamdecaf: I added a struct and parse method for this.

adamdecaf:
I've added `ParsePaymentInformation` onto `BatchENR` and a related struct for all the parsable fields.

batchENR.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, entry := range batch.Entries {
+		if err := entry.Validate(); err != nil {
+			return err
+		}
+
+		if entry.Amount != 0 {
+			msg := fmt.Sprintf(msgBatchAmountZero, entry.Amount, "ENR")
+			return &BatchError{BatchNumber: batch.Header.BatchNumber, FieldName: "Amount", Msg: msg}
+		}
+
```
adamdecaf, line 54: This stuff all needs cleaned up after the Addenda refactor is merged.

wadearnold (approved):

Issue #286, opened 2018-10-04
build: add code coverage to travis-ci steps


adamdecaf:
Pull Requests should have a code coverage bot telling us about coverage percent. Otherwise it's hard to keep good coverage overtime. 

Example without: https://github.com/moov-io/ach/pull/285
Example with: https://github.com/moov-io/auth/pull/33

adamdecaf:
Is it something with my account? @wadearnold do coverage reports show up on this repo for your PR's?

Mine (no coverage report): https://github.com/moov-io/ach/pull/360
Brooke's (with coverage report): https://github.com/moov-io/ach/pull/358

Edit: Wade's last PR on this repo has a report (Sept): https://github.com/moov-io/ach/pull/267

Both PR's were opened today. Mine never seem to have a coverage bot though. 

Switch to codecov?

PR #364, opened 2018-11-08
build: use codecov for coverage analysis

adamdecaf:
goveralls [isn't working for my account](https://github.com/moov-io/ach/issues/286#issuecomment-436463358), so let's replace it with a service that works.

Fixes: https://github.com/moov-io/ach/issues/286


wadearnold (approved):
Do we need to update the coverage badge on the readme in a way that shows the code coverage of the project?

adamdecaf:
Good call. Updated!

Issue #366, opened 2018-11-09
Change TraceNumber and OriginalTrace to strings to prevent errors


adamdecaf:
(The original issue was https://github.com/moov-io/ach/issues/236.)

`TraceNumber` (and likely other fields) have known edge cases and bugs.

- `TraceNumber` is a 15 digit value whose prefix is an ABA routing number.
   - These ABA routing numbers often have a leading zero.
- `int` in Go is only defined: ["int is a signed integer type that is at least 32 bits in size"](https://golang.org/pkg/builtin/#int)

We should change this type to be a string (perhaps a TraceNumber type) to prevent both bugs. This is a breaking change however, so we should release it sooner rather than later.

We should also scan the other fields (`int` types) for similar problems.

adamdecaf:
For reference, here's all other `int` types aside from `TraceNumber` and `OriginalTrace`: 

```Go
	AddendaRecordIndicator             int // 0 or 1
	AddendaRecords                     int // count 
	Amount                             int // non-negative int
	BatchCount                         int // count
	BatchNumber                        int // count 
	BlockCount                         int // count 
	EntryAddendaCount                  int // count 
	EntryDetailSequenceNumber          int // non-negative int
	EntryHash                          int // sum of ABA's 
	ForeignExchangeReferenceIndicator  int // fixed set of values (0, 1, 2)
	ForeignPaymentAmount               int // non-negative int
	OriginatorStatusCode               int // fixed set of values (0, 1, 2)
	SequenceNumber                     int // non-negative int
	ServiceClassCode                   int // fixed set of values
	TotalCreditEntryDollarAmountInFile int // sum
	TotalDebitEntryDollarAmountInFile  int // sum 
	TransactionCode                    int // fixed set of values
```

adamdecaf:
Going by that table I don't think we need to change any other values. Double checking with @bkmoovio and @wadearnold

bkmoovio:
Remind me why we are doing this again?  If I recall you were running a 32 bit system?  Didn’t you put a check in that says we don’t support 32 bit systems?  Just trying to remember the issue?

adamdecaf:
Yes, but the 32bit issue was a problem of overflow.

There's still the problem of any integer type truncating leading 0's from routing numbers.

bkmoovio:
Which routing numbers are not string?

adamdecaf:
TraceNumber (and OriginalTrace) have routing numbers as their prefix.

[...]

PR #367, opened 2018-11-09
Change TraceNumber and OriginalTrace to strings

adamdecaf:
Having these values as `int`'s exposes bugs (leading zero gets truncated). 

Issue: https://github.com/moov-io/ach/issues/366


adamdecaf:
I also scanned through the remaining `int` types to see if others need changed: https://github.com/moov-io/ach/issues/366#issuecomment-437486233

PR #369, opened 2018-11-13
remove crude BatchENR fuzz test

adamdecaf:
This should be done properly with go-fuzz.


adamdecaf:
@bkmoovio ran into errors after pulling down this code. IMO we should run this as another fuzzer not as part of unit tests.

bkmoovio (approved):

Issue #347, opened 2018-11-02
Support XCK - Destroyed Check Entry


wadearnold:
XCK - Destroyed Check Entry - This application can be utilized by a collecting institution for the collection of certain checks when those checks have been destroyed.

- [ ] Library should support the reading and writing of XCK records.
- [ ] Document the XCK record format on the documentation website. 
- [ ] Create a test example for the reading and writing of XCK records

PR #373, opened 2018-11-14
Support TRC- Single check   truncated entries

bkmoovio:
I created a new issue for SEC Code TRX

PR #374, opened 2018-11-14
Support xck   destroyed check entry #347

bkmoovio:
#347


wadearnold (changes requested):
Can you also update the Changelog markdown for the unreleased version 0.4.1. Otherwise this looks perfect!

wadearnold (dismissed):
This pull request for XCK has test cases for TRC. Should those be in #373 ? Can you update the Changelog for XCK support as well?

bkmoovio:
#347

PR #378, opened 2018-11-19
all: mention TraceNumberField() in TraceNumber struct members

adamdecaf:
This should point users towards getting the proper value rather than using the int (which has known bugs).

Issue: https://github.com/moov-io/ach/issues/366

### DIFF
--- a/addenda02.go
+++ b/addenda02.go
@@ -45,6 +45,8 @@ type Addenda02 struct {
 	// TerminalState Identifies the state in which the electronic terminal is located
 	TerminalState string `json:"terminalState"`
 	// TraceNumber Standard Entry Detail Trace Number
+	//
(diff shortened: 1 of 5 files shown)


adamdecaf:
Thoughts @wadearnold ? To merge this instead of https://github.com/moov-io/ach/pull/367

wadearnold:
I could go either with this or #367. Both require you to use the TraceNumberField() function to get a properly formated traceNumber returned. As a string or an int if you set trace number to 1234 your going to get a malformed traceNumber returned without padding. 

I guess I would error on #367 because if I send it a string I should get a string back. If I send it an int I kind of want an int back.

adamdecaf:
Yea. I agree. I'm going to cherry-pick this into that branch so we can merge.

PR #388, opened 2018-12-04
batch: tabulate on build/create calls

adamdecaf:
This change gets rid of the requirements to provide a `batchControl` and/or `fileControl` objects in the JSON for an `ach.File` object. 

Part of: https://github.com/moov-io/moov-io/pull/12

### DIFF
--- a/file.go
+++ b/file.go
@@ -28,7 +28,7 @@ const (
 
 // Errors strings specific to parsing a Batch container
 var (
-	msgFileCalculatedControlEquality = "calculated %v is out-of-balance with control %v"
+	msgFileCalculatedControlEquality = "calculated %v is out-of-balance with file control %v"
 	// specific messages
 	msgRecordLength      = "must be 94 characters and found %d"
 	msgFileBatchOutside  = "outside of current batch"
@@ -118,11 +118,12 @@ func FileFromJSON(bs []byte) (*File, error) {
 
 	// Read FileHeader
 	header := fileHeader{
-		Header: NewFileHeader(),
+		Header: file.Header,
 	}
 	if err := json.NewDecoder(bytes.NewReader(bs)).Decode(&header); err != nil {
 		return nil, fmt.Errorf("problem reading FileHeader: %v", err)
 	}
+	file.Header = header.Header
 
 	if !file.IsADV() {
 		// Read FileControl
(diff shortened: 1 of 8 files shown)


file_test.go
```diff
@@ -422,15 +430,15 @@ func TestFile__readFromJson(t *testing.T) {
 	}
 	batch := file.Batches[0]
 	batchControl := batch.GetControl()
-	if batchControl.EntryAddendaCount != 2 {
+	if batchControl.EntryAddendaCount != 1 {
```
adamdecaf, line 433: @bkmoovio Shouldn't this be `0`? The `ppd-valid.json` file has no addenda records.

adamdecaf: I'm not sure why, but `Batch.build` calls `entryCount++` (which is now `1 + entry.addendaCount()`). That doesn't seem right?

bkmoovio: Hi Adam, 

EntryAddendaCount in the Company/Batch Control Record and File Control Record can be confusing as it is not the AddendaCount for an Entry.   It is  (for reference page OR108) the tally for each Entry Detail Record processed and each Addenda Record processed, within either the batch or file, as appropriate.   For a BatchPPD with one EntryDetail , and no Addenda05, EntryAddendaCount  should be 1,  If the one EntryDetail has an Addenda05, EntryAddendaCount would be 2.  Hopefully that helps . (Clear as Mud)

For the second question:

Prior to removing addendumer change, the code set entryCount as follows:

for i, entry := range batch.Entries {
		entryCount = entryCount + 1 + len(entry.Addendum)....

going forward I had added +1 to entryCount for each entry (EntryDetail) + each Addenda property if the property was defined. Addenda02, len(Addenda05), Addenda98, Addenda99.  it looks like you created a function for that which is great.  I should have.

bkmoovio: Regarding ++,

https://github.com/golang/go/issues/21263

To funny... :-)

adamdecaf: Ah, okay. Those names are confusing but it makes sense.

file.go
```diff
@@ -177,7 +179,7 @@ func (f *File) setBatchesFromJSON(bs []byte) error {
 	if err := json.Unmarshal(bs, &batches); err != nil {
 		return err
 	}
-	// Clear out any nil batchs
+	// Clear out any nil batchess
```
bkmoovio, line 182: batches

bkmoovio (approved):

file_test.go
wadearnold: well that's a good idea.

Issue #391, opened 2018-12-04
Add const values for BatchHeader.ServiceClassCode and BatchControl.ServiceClassCode


bkmoovio:
We have a fair amount of hard-coded values, which aren't that friendly for developers. If we provide names it should help to understand them.

Example: BatchHeader.ServiceClassCode: 200, 220, 225, 280

Identifies the general classification of dollar entries to be exchanged (i.e. debit and/or credit entries) 
 –200   ACH Entries Mixed Debits and Credits 
 –220   ACH Credits Only 
 –225   ACH Debits Only 
 –280   ACH Automated Accounting Advice

Create developer friendly meaningful constants for these values and modify the code to reference the constants.

PR #394, opened 2018-12-06
Add const values for batch header.service class code

bkmoovio:
#391

### DIFF
--- a/advBatchControl.go
+++ b/advBatchControl.go
@@ -18,7 +18,7 @@ type ADVBatchControl struct {
 	ID string `json:"id"`
 	// RecordType defines the type of record in the block.
 	recordType string
-	// AutomatedAccountingAdvices: 280
+	// Constant AutomatedAccountingAdvices: 280
 	// Same as 'ServiceClassCode' in BatchHeaderRecord
 	ServiceClassCode int `json:"serviceClassCode"`
 	// EntryAddendaCount is a tally of each Entry Detail Record and each Addenda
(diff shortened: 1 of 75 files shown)


advBatchControl.go
```diff
@@ -18,9 +18,7 @@ type ADVBatchControl struct {
 	ID string `json:"id"`
 	// RecordType defines the type of record in the block.
 	recordType string
-	// ServiceClassCode ACH Mixed Debits and Credits ‘200’
-	// ACH Credits Only ‘220’
-	// ACH Debits Only ‘225'
+	// Constant AutomatedAccountingAdvices: 280
```
wadearnold: Should this be commented out?

advBatchControl.go
```diff
@@ -60,7 +58,8 @@ func (bc *ADVBatchControl) Parse(record string) {
 
 	// 1-1 Always "8"
 	bc.recordType = "8"
-	// 2-4 This is the same as the "Service code" field in previous Batch Header Record
+	// 2-4 This is the same as the "Service code" field in previous Batch Header Record -
+	// AutomatedAccountingAdvices: 280
```
wadearnold: Does this comment need to be here? Not sure what it tells me.

advBatchControl_test.go
```diff
@@ -49,12 +49,12 @@ func BenchmarkMockADVBatchControl(b *testing.B) {
 
 // TestParseADVBatchControl parses a known Batch ControlRecord string.
 func testParseADVBatchControl(t testing.TB) {
-	var line = "822500000100053200010000000000000001050000000000000000000000T-BANK             076401250000001"
+	var line = "828000000100053200010000000000000001050000000000000000000000T-BANK             076401250000001"
```
wadearnold, line 52: Nice! Good catch!

wadearnold (approved):

advBatchControl.go
bkmoovio: Ok I'll remove.  No problem!

advBatchControl.go
bkmoovio: I'll remove!

bkmoovio: since this in adv control, I'll have it be this

	// This should be the same as BatchHeader ServiceClassCode for ADV: AutomatedAccountingAdvices.

Issue #392, opened 2018-12-04
Add const values for BatchHeader.StandardEntryClassCode


bkmoovio:
We have a fair amount of hard-coded values, which aren't that friendly for developers. If we provide names it should help to understand them.

Example: BatchHeader.StandardEntryClassCode:  ACK, ADV, ARC, ATX, BOC, CCD, CIE, COR, CTX, DNE, ENR, IAT, MTE, POP, POS, PPD, RCK, SHR, TEL, TRC, TRX, WEB

Create developer friendly meaningful constants for these values and modify the code to reference the constants.

PR #395, opened 2018-12-07
Add const values for batch header.standard entry class code

bkmoovio:
#392


wadearnold (approved):
Looks great. Thanks for adding the additional test coverage as well.

adamdecaf (approved):

Issue #412, opened 2018-12-14
server: unit test to create all repo test files against ACH Server


adamdecaf:
The idea is to scrape up all ACH files in our repository and try creating them against our API. Let's try both in the ACH text format and JSON.

bkmoovio:
#412

PR #416, opened 2018-12-18
Unit test to create all repo test files against ach server

bkmoovio:
#412


file.go
```diff
@@ -166,20 +166,39 @@ func (f *File) UnmarshalJSON(p []byte) error {
 	return errors.New("json struct tag unmarshal is deprecated, use ach.FileFromJSON instead")
 }
 
+// ACHFileToJSON returns an ACH file in JSON format from ACH.File.
+func (f *File) ACHFileToJSON() ([]byte, error) {
```
adamdecaf, line 170: I don't see why this method is needed. Callers can just issue `json.Marshal(fileJSON)` themselves.

iatBatch.go
```diff
@@ -45,6 +46,23 @@ func NewIATBatch(bh *IATBatchHeader) IATBatch {
 	return iatBatch
 }
 
+// UnmarshalJSON un-marshals JSOn IATBatch
+func (batch *IATBatch) UnmarshalJSON(p []byte) error {
```
adamdecaf, line 50: Can we add a `if batch == nil` check and set the value? `*batch = NewIATBatch()`

Otherwise this method would fail on `var batch *IATBatch` I believe.

server/test/createFile_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"testing"
+
+	"github.com/moov-io/ach"
+)
+
+// TestCreateFile Tests Creating an ACH File From JSON
+// These tests are not designed to have a failure but can be used as TEST for JSON formatting for an SEC Code
+func TestCreateFile(t *testing.T) {
+	createFileError := func(path string, msg string) {
+		_, err := os.Open(path)
```
adamdecaf, line 21: What if instead we read the existing `*.ach` files in each `test/*-read/` directory, convert those to JSON in memory and then check? We wouldn't need to keep the JSON files in this repository or up to date in future changes.

bkmoovio: Sure we can do that.  I was thinking we would want to give users sample json files for each SEC code.

bkmoovio: Im not sure the point of this issue then because we know we support creating files from json.

bkmoovio: Im not sure the point of this issue then because we know we support creating files from json.

file.go
bkmoovio: That’s correct, but I wanted in the library.

adamdecaf: Could we name it `ToJSON` then?

server/test/createFile_test.go
adamdecaf: If we want the JSON files to live in the repo then can we generate them as part of `make` and put them in an `examples/json/` directory.

file.go
bkmoovio: Yes, I think so, if that is preferable.

iatBatch.go
bkmoovio: I will check.  I didn’t think it compiled when I had the nil check.    I’ll review.

server/test/createFile_test.go
bkmoovio: So you want to automatically create the json files based on the *.ach files in the ach test directory.  That’s fine.

file.go
bkmoovio: I actually like the name as is.  I prefer to leave it.

iatBatch.go
bkmoovio: that was in file.go, that I removed the nil check.

server/test/createFile_test.go
bkmoovio: Actually is there real value to that?   How often do we anticipate the *.ach files changing?  If we have a make file that creates the json files do the json files get committed to the repo?

bkmoovio: I certainly am ok doing it.  I just want to make sure, we have the examples.

bkmoovio: In the interim I'll look at putting this in the test\server-example directory or the examples directory.  From there I can automate it...

file.go
adamdecaf: Let's rename this to `ToJSON`. The current name is wordy.

```Go
file.ACHFileToJson()

// vs

file.ToJSON() 
```

server/test/createFile_test.go
adamdecaf: I'd like it to be automated as much as possible that way we don't have to maintain all these JSON files going forward. Having a unit test write them to disk is fine, but could introduce noisy diffs in the future.

bkmoovio: Ok got it.

file.go
adamdecaf, line 170: I don't see the need for this method or the wordy naming, but let's call it `ToJSON` with the following body:

```Go
func (f *File) ToJSON() ([]byte, error) {
	return json.Marshal(f)
}
```

Edit: The standard Go way to name methods like this is `ToXXX`: 

```
$ go doc net IP 
...
func (ip IP) To16() IP
func (ip IP) To4() IP
```

adamdecaf (changes requested):

file.go
bkmoovio: I’m going to use MarshalIndent in the caller

Issue #275, opened 2018-10-01
more durable storage ? 


adamdecaf:
**What were you trying to do?**
I was trying to manipulate a file through `ach` service restarts, but after restart the file is missing.

**What did you expect to see?**
The file exist throughout restarts (i.e. deploys).

adamdecaf:
We could easily mount a volume and write JSON flat files. Add a worker to delete files after N days?

adamdecaf:
Thinking about this.. Is a good usecase for nosql this type of storage? I don't think we want to get into SQL persistence for these objects.

There's a sqlite extension we could look at: https://www.sqlite.org/json1.html

Also, Mongo has gotten a lot better in the 3.x series. We might need to implement as it's still a popular db.

wadearnold:
A three-node mongo cluster seems reliable now. A hosted version is even better. I think for our current use case a single-node mongo db would work great. We can build out our reliability story later. I also agree that its a common data store.

zbruhnke:
Could be worth just using CouchDB for this especially if planning on using a single node cluster to start as its more suited to single node clusters and can easily scale well with replication later ... just thinking out loud but seems more than apt for this use case - It's also Apache 2.0 licensed which is nice considering the licenses on all the Moov repos

wadearnold:
@zbruhnke Thanks for the feedback. I like the idea of reliable storage that still "works on my computer" is ideal. One of the main design goals of Moov is to be able to: 

> Develop locally and deploy globally

This comes from 15 years of experience in banking where I needed a VPN and ISDN line to the IBM mainframe to run a test. GAH! 

All of our projects have been designed first with straight SQL against sqlite for in-memory local development. The GoLang interface for storage should allow us to add any other persistence layer easily.

PR #432, opened 2019-01-08
server: add TTL config for expiring old Files from in-mem repository

adamdecaf:
We're seeing lots of test files accumulate and so we should provide a way to expire old files.

Issue: https://github.com/moov-io/ach/issues/275


adamdecaf:
My plan is to deploy with a 4 hour TTL on production: 

```yaml
env:
  - name: ACH_FILE_TTL
    value: 240m # 4 hours
```

bkmoovio (approved):

PR #450, opened 2019-01-21
internal: use filepath.Join for cross-platform paths in codegen

adamdecaf:
Noticed during https://github.com/moov-io/gl/pull/3


adelelopez (approved):
Nice catch!

PR #452, opened 2019-01-22
Proof of concept for type-based error handling

adelelopez:
Here's a potential way we could improve error handling. Each type of error is now literally its own type implementing the `Error` interface. A helper function is provided (and which would likely be moved into `moov-io/base` if we decided to use this) which lets us easily check whether the error list returned has the specific error we are expecting. This is demonstrated in the updated tests.

The downside of this method is that each new kind of error has to have its own type and `Error()` function, along with a constructor if the error has details. This can be a bit cumbersome.

The upside is that using errors becomes significantly easier. It also will make localization easier down the road, since all the error messages will be in one place and can depend on a locale configuration variable.

Let me know if this seems reasonable to you, or if you have any other kinds of comments or suggestions!

### DIFF
--- a/fileErrors.go
+++ b/fileErrors.go
@@ -12,8 +12,26 @@ import (
 	"github.com/moov-io/base"
 )
 
-// ErrFileTooLong is the error given when a file exceeds the maximum possible length
-var ErrFileTooLong = errors.New("file exceeds maximum possible number of lines")
+var (
+	// ErrFileTooLong is the error given when a file exceeds the maximum possible length
+	ErrFileTooLong = errors.New("file exceeds maximum possible number of lines")
+	// ErrFileHeader is the error given if there is the wrong number of file headers
+	ErrFileHeader = errors.New("none or more than one file headers exists")
+	// ErrFileControl is the error given if there is the wrong number of file control records
+	ErrFileControl = errors.New("none or more than one file control exists")
+	// ErrFileEntryOutsideBatch is the error given if an entry is outside of a batch
+	ErrFileEntryOutsideBatch = errors.New("entry outside of batch")
+	// ErrFileAddendaOutsideEntry is the error given if an addenda is outside of an entry
+	ErrFileAddendaOutsideEntry = errors.New("addenda outside of entry")
+	// ErrFileBatchControlOutsideBatch is the error given if a batch control record is outside of a batch
+	ErrFileBatchControlOutsideBatch = errors.New("batch control outside of batch")
+	// ErrFileBatchHeaderInsideBatch is the error given if a batch header record is inside of a batch
+	ErrFileBatchHeaderInsideBatch = errors.New("batch header inside of batch")
+	// ErrFileADVOnly is the error given if an ADV only file has a non-ADV batch
+	ErrFileADVOnly = errors.New("file can only have ADV Batches")
+	// ErrFileIATSEC is the error given if an IAT batch uses the normal NewBatch
+	ErrFileIATSEC = errors.New("IAT Standard Entry Class Code should use iatBatch")
+)
 
 // RecordWrongLengthErr is the error given when a record is the wrong length
 type RecordWrongLengthErr struct {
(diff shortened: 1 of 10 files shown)


fileErrors.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/moov-io/base"
+)
+
+// FileTooLongErr is the error given when a file exceeds the maximum possible length
+type FileTooLongErr string
```
adamdecaf: If we wanted to cut down on the error types what about:

```go
type FileErr struct {
	Description string
}
func (e FileErr) Error() string {
	return fmt.Sprintf("FileErr: %s", e.Description)
}

func NewFileErr(desc string) FileErr {
	return FileErr{desc} 
}
```

reader_test.go
```diff
@@ -2159,18 +2148,8 @@ func testACHFileTooLongErr(t testing.TB) {
...
-				if e.Msg != msgFileTooLong {
-					t.Errorf("%T: %s", e, e)
-				}
-			}
-		} else {
-			t.Errorf("%T: %s", el, el)
-		}
-	} else {
-		t.Errorf("No error even though file was too long")
+	if !Has(err, FileTooLongErr("file too long")) {
```
adamdecaf: This defiantly cleans up the test cases!

fileErrors.go
adelelopez: We could do that, but then we would need to change how we check what kind of error it is. This PR's implementation checks for the same type. We would have to adjust that to additionally check that the details (e.g. `Description` in this case) match. So that basically takes us back to just matching the error strings to do this check, which is more fragile. Of course we could save those strings as constants like we do now, but in that case this whole thing seems kinda redundant...

Maybe instead of using separate types for simple errors (like `FileTooLongErr`), we could do something like this: `var FileTooLongErr error = errors.New("File exceeds maximum possible length")`. I think the `Has` function can be adjusted to deal with this kind of error as well. Then we could just drag out the whole type thing only when we want the error to carry extra information (like `RecordWrongLengthErr`).

reader_test.go
bkmoovio: I like how this cleans up the tests!

fileErrors.go
bkmoovio: If Has function can be adjusted this might make everything easier 

Maybe instead of using separate types for simple errors (like FileTooLongErr), we could do something like this: var FileTooLongErr error = errors.New("File exceeds maximum possible length"). I think the Has function can be adjusted to deal with this kind of error as well. Then we could just drag out the whole type thing only when we want the error to carry extra information (like RecordWrongLengthErr).

adamdecaf: I think `FileTooLongErr := errors.New("...")` is a better way for most error values. Your point on i18n is important too in that having a consolidated place to change in the future will simplify that work.

adelelopez: Alright, sounds good! I'll put up a version with this change.

reader_test.go
bkmoovio: I'm ok with the changes @Adam?

For the error on testACHFileRead3,  It's a malformed file (intentionally).  It looks like I had it checking three different errors in the file one from read() and two from File.Validate(), so that can be done or at this point removing the test is fine as it shouldn't decrease reader.go test coverage.

I haven't looked into TestFileEntryDetail so if you have time, you can but I'm guessing it probably needs checks for the 3 errors EntryDetail Record Error, FileHeader Record Error and FileControl Record Error. It seems the EntryDetail Record  - line := ed.String() is coming up short.

The EntryDetail error needs to be a FileError for msgRecordLength .

reader_test.go
```diff
@@ -461,8 +425,6 @@ func testFileBatchHeaderErr(t testing.TB) {
 					t.Errorf("%T: %s", e, e)
 				}
 			}
-		} else {
-			t.Errorf("%T: %s", el, el)
```
adelelopez, line 465: This test was already broken, but the changes in this PR made it not pass. So I took this case out so that it would pass again. This will get fixed when I make the equivalent changes to the batch errors.

adamdecaf (approved):

Issue #469, opened 2019-02-12
RecordType tests and other tests that need to be checked


adelelopez:
While working on https://github.com/moov-io/ach/pull/468, I came across several tests for which it was not readily apparent whether they were working as intended. In order to reduce the scope of that PR, I simply labeled these tests with a `TODO` indicating that they should be reexamined.

PR #471, opened 2019-02-15
Fix various ach tests

adelelopez:
This fixes a bug I introduced where no error is returned for the record type checks (see https://github.com/moov-io/ach/issues/469). 

There are also fixes of various bugs that I came across, including:

- reading a file from JSON produced an incorrect file
- the originator status code was defaulted to 0, which is for ADV batches only
- MTE batch identification numbers were not being validated
- some tests were not creating batches in the correct manner


.codecov.yml
```diff
@@ -27,3 +27,4 @@ comment:
 
 ignore:
   - "*Errors.go"  # ignore error files
+  - "file.go"
```
adamdecaf: What's the reason for ignoring this file?

adelelopez: I just did this because I don't have time to write json tests for every batch type, and I'm not sure how worth doing that is.

adelelopez: I just did this because I don't have time to write json tests for every batch type, and I'm not sure how worth doing that is.

file.go
```diff
@@ -175,10 +175,85 @@ func (f *File) setBatchesFromJSON(bs []byte) error {
...
+		}
+
+		if err := batch.build(); err != nil {
+			return batch.Error("Invalid Batch", err, batch.Header.ID)
+		}
+
+		// Attach a batch with the correct type
+		switch batch.Header.StandardEntryClassCode {
+		case ACK:
+			f.Batches = append(f.Batches, &BatchACK{batch})
```
adelelopez: Every case here needs to get tested to pass code coverage. So a test JSON file for each batch type will need to be tested. That's why I just removed this file from code coverage.

.codecov.yml
adamdecaf: yea, I thought we already had these JSON tests.

file.go
bkmoovio: Creating a JSON file for each batch type would seem to be a separate issue(s).

.codecov.yml
adelelopez: Anyway, did a refactor so that this point is now moot

bkmoovio (approved):

PR #491, opened 2019-02-28
build: remove vendor and tidy dependencies 

adamdecaf:
The renovate bot isn't updating `vendor/`, so the dependencies in there are out of sync with `go.mod`. This can cause someone building with `vendor/` to get a different result than someone with Go Modules.


adamdecaf:
cc @bkmoovio Could you try this branch out?

PR #496, opened 2019-03-01
Check for proper entryHash overflow

bkmoovio:
Add entryHash test

### DIFF
--- a/test/fuzz-reader/reader_test.go
+++ b/test/fuzz-reader/reader_test.go
@@ -13,29 +13,29 @@ import (
 )
 
 func TestCorpusSymlinks(t *testing.T) {
-
 	// avoid symbolic link error on windows
-	if runtime.GOOS != "windows" {
-		fds, err := ioutil.ReadDir("corpus")
-		if err != nil {
-			t.Fatal(err)
-		}
-		if len(fds) == 0 {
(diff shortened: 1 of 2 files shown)


adamdecaf (approved):
Could we add a comment to the test about how we're testing the truncation of the checksum?

test/fuzz-reader/reader_test.go
```diff
@@ -8,29 +8,34 @@ import (
...
-	}
 
-	for i := range fds {
-		if fds[i].Mode()&os.ModeSymlink != 0 {
-			if path, err := os.Readlink(filepath.Join("corpus", fds[i].Name())); err != nil {
-				t.Errorf("broken symlink: %v", err)
-			} else {
-				if _, err := os.Stat(filepath.Join("corpus", path)); err != nil {
+	// avoid symbolic link error on windows
+	if runtime.GOOS != "windows" {
```
adamdecaf: Could we just skip the test early?

```go
if runtime.GOOS == "windows" {
	t.Skip("git symlinks aren't supported on windows")
} 
```

Also, I was going to put this in a different PR as the other change isn't related to this test failure.

bkmoovio: It's fine if you want to change where the skip is under another PR.

Issue #529, opened 2019-04-30
File Creation - Merge Files


wadearnold:
A utility function should exist that takes an array of `ACH.Files` and merges them into one or many files. If the merged file exceeds the NACHA record limit limitation of 10k entries and the additional file should be created.

adamdecaf:
```go
// MergeFiles is a helper function for consolidating an array of ACH Files into as few files
// as possible. This is useful for optimizing cost and network efficiency.
//
// Per NACHA rules files must remain under 10,000 lines (when rendered in their ASCII encoding)
//
// File Batches can only be merged if they are unique and routed to and from the same ABA routing numbers.
func MergeFiles(files []*File) ([]*File, error) { ... }
```

@wadearnold and/or @bkmoovio would you have other comments to add here? 

I'm going to use some of paygate's [`mergeTransfer` logic](https://github.com/moov-io/paygate/blob/master/file_transfer_async.go#L529) and [line count wrapper](https://github.com/moov-io/paygate/blob/master/file_transfer_async.go#L821) for this.

bkmoovio:
I think it’s good.  Wade can probably chime in on if any specific additional requirements are needed.  It’s great if you can leverage what you did in paygate.

PR #534, opened 2019-05-23
 files: support arbitrary merging of ACH files

adamdecaf:
This pull request implements merging of arbitrary ACH files based on their `FileHeader` structs. The code optimizes for the fewest files by ABA routing information, but does not consolidate Batches with the same SEC codes. 

Files provided are assumed to be valid and under NACHA's 10k line limit and this function will always return valid files when provided with valid files. An error will be returned on invalid input files. 

Issue: https://github.com/moov-io/ach/issues/529


wadearnold (approved):
Let's make sure that we add this to the readme somewhere and also as a new feature. I could see people using the library just for this functionality.

PR #535, opened 2019-05-23
batch: only check DNE specifics if the SEC code is DNE

adamdecaf:
This came up when a PPD prenote was created, but this check was too
broad and returned an error incorrectly.


wadearnold (approved):

bkmoovio:
It looks good to me!

adamdecaf:
I cherry-picked this into a v1.0.1 release

Issue #531, opened 2019-04-30
File Creation - Balanced Offset


wadearnold:
A Balanced Offset is a Batch with Entry Detail that creates a balanced file. The offset is not a single net new entry. If there are debits, there is a credit offset. If there are credits, there is a debit offset. They are mutually exclusive. Offset files are used to offset transactions from a single account inside of the ODFI. The Offset is the last record which settles the funds into a single ODFI account. 

The process is to create a Batch Header with an Entry Detail providing the offsetting account information, amount, and summing all previous entries. 

The proposed solution is to add a function that would summarize the offset when a file was being created. 

Pseudocode 

```golang    
    // … assume all batches and corresponding entries have been added to the File 

    os_bh := NewBatchHeader()
    os_bh.StandardEntryClassCode = CCD
    os_bh.CompanyName = "ACME Corporation"
    os_bh.CompanyIdentification = "121042882"
    os_bh.CompanyEntryDescription = “OFFSET"
    os_bh.ODFIIdentification = “12104288”
   // All default properties will be overwritten when calculate

    os_entry := NewEntryDetail()
    os_entry.SetRDFI("121042882")
    os_entry.DFIAccountNumber = "123456789"
    os_entry.IndividualName = “OFFSET”
    // All default properties will be overwritten when calculated

    file.BalancedOffset(os_bh, os_entry)
    if err := file.Create(); err != nil {
        t.Fatalf("Unexpected error building file: %s\n", err)
    }
 ```

bkmoovio:
With regard to: 'and summing all previous entries'

Should be able to use batch.GetControl().TotalDebitEntryDollarAmount  or  batch.GetControl().TotalCreditEntryDollarAmount, which ever is greater than zero to determine the offset amount.

If both are greater than zero we are unable to have one offset entry so return an error. 

If it is a new batch, create the BatchHeader data, create the Batch based on SEC Code, add the entry, add the Batch to the existing file (consider line count 10000) followed by file.Create() to create the new file.  

If it is a new entry in the batch, add the entry to the batch and call Batch.build(), followed by file.Create() to create a new file.  

Delete the old file?

adamdecaf:
Sounds good. `file.Create()` doesn't make a new file, so there wouldn't be anything to cleanup.

bkmoovio:
Thanks Adam, that's great!

Issue #528, opened 2019-04-30
File Creation - Segment Debit/Credit 


wadearnold:
File creation should default to NACHA rules by supporting both credits and debits in the same ACH file for type codes. Some legacy systems require the file to only contain Debits (deposit) or Credits (withdraws) in a file. Segmenting Debits and Credits should be a configuration for the client. 

The proposed configuration would iterate through the file and write all batches with a ServiceClassCode of `220` into a credits file and `225` into a debits file. Service class code `200 would need to create new batches in the appropriate debits or credits files.

Rather than a configuration, this could be a post file creation process to segment the debits and credits that would take an input of an  `ACH.File` being created and return two `ACH.Files` that have been segmented. 

We should be aware of how to deal with the Balanced Offset Record when segmenting files.

PR #537, opened 2019-05-29
batch: add WithOffset to support balanced offset records

adamdecaf:
Balanced offset records sum a File's Debits or Credits to zero in a single EntryDetail within a Batch. These are used to offset transactions within an ODFI.

Issue: https://github.com/moov-io/ach/issues/531


file_test.go
```diff
@@ -1050,3 +1050,103 @@ func TestFile__RemoveBatch(t *testing.T) {
...
+	bh.CompanyName = "ACME Corporation"
+	bh.CompanyIdentification = "121042882"
+	bh.CompanyEntryDescription = "OFFSET"
+	bh.ODFIIdentification = "12104288"
+
+	ed := NewEntryDetail()
+	ed.SetRDFI("121042882")
+	ed.DFIAccountNumber = "123456789"
+	ed.IndividualName = "OFFSET"
+	ed.TransactionCode = CheckingCredit
```
adamdecaf: I don't like how we needed to specify Checking(Credit/Debit) as that defeats the purpose of this method (to automate determining that in the method), but I'm unsure how we could pick between Checking and Savings as we don't know the ODFI account types. Thoughts?

file.go
```diff
@@ -653,3 +653,61 @@ func (f *File) createFileADV() error {
...
+		return fmt.Errorf("CalculateBalancedOffset: batch create: %v", err)
+	}
+	f.AddBatch(batch)
+
+	// Perform some final checks on the file
+	if err := f.Create(); err != nil {
+		return fmt.Errorf("CalculateBalancedOffset: final file create: %v", err)
+	}
+	if n, err := lineCount(f); err != nil || n > NACHAFileLineLimit {
+		return fmt.Errorf("CalculateBalancedOffset: NACHA file length limit reached (lines=%d): error=%v", n, err)
```
adamdecaf: Should we call `RemoveBatch` within the error branch here? The file isn't valid per NACHA rules with the offset record added.

file_test.go
bkmoovio: Good question, I believe that was part discussion we had initially,  let the caller do the work instead of creating a function do it.  So the caller creates a batch with the entry detail record that is the offset and adds the batch to file similar to any other batch being added to a file.

file.go
bkmoovio: What is the error?

bkmoovio (commented):
Should that be BatchControl instead of FileControl?

file.go
```diff
@@ -653,3 +653,61 @@ func (f *File) createFileADV() error {
...
+//
+// If a file contains Debits the offset record will be a Credit and if a file contains Credits
+// the offset record will be a debit.
+//
+// The ACH File will have its FileControl record retabulated after appending the Batch.
+func (f *File) CalculateBalancedOffset(bh *BatchHeader, ed *EntryDetail) error {
+	if err := f.Create(); err != nil {
+		return fmt.Errorf("CalculateBalancedOffset: initial tabulate: %v", err)
+	}
+	if f.Control.TotalDebitEntryDollarAmountInFile > 0 && f.Control.TotalCreditEntryDollarAmountInFile > 0 {
```
bkmoovio: Should that be BatchControl instead of FileControl or are we presuming only one batch per file?

file_test.go
adamdecaf: Right, but I was hoping the caller didn't need to know about debit vs credit. It sorta defeats the purpose of this helper method if they have to account for it also.

file.go
adamdecaf: That errors when the file is over NACHA's 10k line limit.

file.go
adamdecaf: I thought offset records were for an entire file.

file_test.go
bkmoovio: I agree the helper should make it easier then having the caller set the entryDetail properties, but as the code is now, an entryDetail that is the offset entry is feasible.

file.go
bkmoovio: If it is over the 10K line limit do we have to split the batch(es) into separate files?  If the caller was responsible for the offset, the caller would be responsible for that.  Also if we have a 10K line limit do we need to adjust the following:

	// maxLines is the maximum number of lines a file can have. It is limited by the
	// EntryAddendaCount field which has 8 digits, and the BatchCount field which has
	// 6 digits in the File Control Record. So we can have at most the 2 file records,
	// 2 records for each of 10^6 batches, 10^8 entry and addenda records, and 8 lines
	// of 9's to round up to the nearest multiple of 10.
	maxLines = 2 + 2000000 + 100000000 + 8

file.go
bkmoovio: My understanding was offset records are used in batches, to create a balanced file.

bkmoovio: The current library creates ach files which contain batches with detail entries the caller specifies.  The caller can create an offset detail entry in the batch.  My understanding of this enhancement is to add a helper function to create an offset entry so the caller doesn't have to create the offset entry.

file.go
adamdecaf: I'd say the caller has to be responsible for splitting the Batches into files.

adamdecaf: I forgot that our `Reader` has a `maxLines` variable declared. That's for reading, not writing (or generating) files.

bkmoovio:
Sounds good.

bkmoovio:
That’s correct it is for reading.

batch.go
```diff
@@ -991,3 +1004,103 @@ func (batch *Batch) Equal(other Batcher) bool {
...
+		} else {
+			ed.TransactionCode = SavingsDebit
+		}
+	default:
+		return fmt.Errorf("offset: unknown account type %v", b.offset.AccountType)
+	}
+	ed.RDFIIdentification = b.offset.RoutingNumber[:8]
+	ed.CheckDigit = b.offset.RoutingNumber[8:9]
+	ed.DFIAccountNumber = b.offset.AccountNumber
+	ed.IdentificationNumber = "" // TODO(adam): what to put here?
```
adamdecaf: What should go here? Any ideas?

batch.go
```diff
@@ -7,8 +7,16 @@ package ach
...
+	"fmt"
+	"math/rand"
 	"strconv"
 	"strings"
+	"time"
+	"unicode/utf8"
+)
+
+var (
+	traceNumberSource = rand.NewSource(time.Now().Unix())
```
wadearnold: What is traceNumberSource used for?

batch.go
```diff
@@ -410,7 +423,7 @@ func (batch *Batch) build() error {
 		bcADV.TotalCreditEntryDollarAmount, bcADV.TotalDebitEntryDollarAmount = batch.calculateADVBatchAmounts()
 		batch.ADVControl = bcADV
 	}
-	return nil
+	return batch.upsertOffset()
```
wadearnold: upsert means insert right?

batch.go
```diff
@@ -991,3 +1004,121 @@ func (batch *Batch) Equal(other Batcher) bool {
...
+
+func (b *Batch) upsertOffset() error {
+	if b == nil || b.offset == nil {
+		return nil
+	}
+
+	// remove any Offset records already on the batch
+	addOffset := true // append to EntryAddendaCount
+	for i := 0; i < len(b.Entries); i++ {
+		// TODO(adam): Should we remove this based on checking the last element is
```
wadearnold, line 1020: Can the TODO be moved to a GIthub issue as a question or enhancement

batch.go
```diff
@@ -991,3 +1004,121 @@ func (batch *Batch) Equal(other Batcher) bool {
...
+			} else {
+				b.Control.TotalDebitEntryDollarAmount = 0
+			}
+			// remove the EntryDetail
+			b.Entries = append(b.Entries[:i], b.Entries[i+i:]...)
+			addOffset = false // we're replacing an EntryDetail, not adding
+			i++
+		}
+	}
+	if b.Control.TotalDebitEntryDollarAmount > 0 && b.Control.TotalCreditEntryDollarAmount > 0 {
```
wadearnold: Is this true that an offset record can only have debits or credits but not both? I thought that the customer in #general said she couldn't have both.  Wouldn't we just change the service calls code based on debit, credit, or both? 
200 ACH DEBITS AND CREDITS
220 ACH CREDITS ONLY
225 ACH DEBITS ONLY

batch.go
wadearnold: I don't see this being used for this use case. 
"An internal identification (alphanumeric) that you use to uniquely identify this Entry Detail Record"

batch.go
```diff
@@ -991,3 +1004,121 @@ func (batch *Batch) Equal(other Batcher) bool {
...
+	if n := utf8.RuneCountInString(rtn); n == 10 {
+		return rtn[1:9] // ACH server will prefix with space, 0, or 1
+	}
+	if n := utf8.RuneCountInString(rtn); n != 8 && n != 9 {
+		return ""
+	}
+	return rtn[:8]
+}
+
+func createTraceNumber(odfiRoutingNumber string) string {
```
wadearnold: Don't we already have something that does this ?

batch.go
```diff
@@ -991,3 +1004,121 @@ func (batch *Batch) Equal(other Batcher) bool {
...
+			return SavingsDebit, nil
+		}
+	default:
+		return 0, fmt.Errorf("offset: unknown account type %v", b.offset.AccountType)
+	}
+}
+
+// aba8 returns the first 8 digits of an ABA routing number.
+// If the input is invalid then an empty string is returned.
+func aba8(rtn string) string {
```
wadearnold, line 1097: Do we not have something that does this already?

wadearnold (approved):

batch.go
adamdecaf: We randomly generate a TraceNumber for the offset record. Maybe we should inherit from the surrounding entries? I didn't want to get too far into parsing TraceNumbers and incrementing as we might collide with a 3rd party incrementing scheme.

batch.go
adamdecaf: Nope. `upsert` is to "update or insert". It's an idempotent operation on the Batch that always leaves an offset record as the last EntryDetail.

batch.go
bkmoovio: I think you would have to change the service class code to 200 if it was 220 or 225 and an  offset record is created.  I think if you were going to create offset records for a mixed batch, would the code stop creating offsets after the first offset for credits/debits?  OFFSET can probably be used to determine that, and/or some sorting/ordering by credit/debit.

batch.go
adamdecaf: Done - https://github.com/moov-io/ach/issues/540

batch.go
adamdecaf: I understood our chats and their problem as only allowing debits or credits within a batch and the offset was the opposite to zero out. 

If a Batch can have debits and credits and needs a balancing record that's fine. We can do the math to figure it out still.

batch.go
adamdecaf: I'll leave it as an empty string then.

batch.go
adamdecaf: This was copied from paygate, but I didn't see a function like this in the ACH project.

batch.go
adamdecaf: In paygate yes, but not in ACH. See https://github.com/moov-io/ach/pull/537/files#r290038772

batch.go
wadearnold: I would think the sequence number needs to be increment from what is previously in the Batch? Should we check in batch.build if this is a balanced file and increment it in this section? 
https://github.com/moov-io/ach/blob/21375f23ca5ce5ef72629a3fbb567298cafdccd0/batch.go#L362

batch.go
wadearnold: Thanks for the clarification

batch.go
wadearnold: A batch can have debits and credits and have an offset. I think these are two problems. That customer needs to separate debits and credits into two different files with each having an offset. We have a separate feature request for segmenting debits and credits. 
https://github.com/moov-io/ach/issues/528

wadearnold (changes requested):
I have no idea how we would expose this over an HTTP API. I assume we add it to the GoLang Library first and then add the API second?

batch.go
adamdecaf: Yea. The TraceNumber now comes from previous EntryDetail records.

batch.go
adamdecaf: So a batch with debits and credits can't have an offset record created? It sounds like in that case we should return an error and direct the client to split the file and then call `WithOffset` on each.

adamdecaf:
I'm not sure how we'll expose this over the HTTP API either. I am assuming we'll want to support splitting one file into two (debits and credits) in the API also?

wadearnold:
In a balanced file the offset is not a single net entry.  If there are debits (27), there is a credit offset (22) matching the sum of the debits.  If there are credits (22), there is a debit(27) offset matching the sum of the credits.  They are mutually exclusive.

wadearnold:
> I'm not sure how we'll expose this over the HTTP API either. 

Is a balanced file a different resource or a service on a resource? I see this more as a pipeline decorator to an ACH File. I have a file as a resource and now I want to mutate it in this way. Maybe I post a url of a file at an endpoint that returns a balanced resource?  



> I am assuming we'll want to support splitting one file into two (debits and credits) in the API also?

I see these as two different issues but with a similar api architecture.

adamdecaf:
> In a balanced file the offset is not a single net entry. If there are debits (27), there is a credit offset (22) matching the sum of the debits. If there are credits (22), there is a debit(27) offset matching the sum of the credits. They are mutually exclusive.

Ohhh! Each batch needs to have two offset records? One debit and one credit? That makes sense.

adamdecaf:
What about endpoints like this? They won't store any data, but transform the file and return the result.

`POST /files/:id/balance` (We could offer parameters here if needed.)
`POST /files/:id/split`  (Assumes split into two files, one debit only and one credit only)

wadearnold:
> Ohhh! Each batch needs to have two offset records? One debit and one credit? That makes sense.

My understanding: 

- if all of the entries are credit you would only have a single debit offset. 
- If all the entries are debit you would have a single credit offset. 
- If you have a mixed debit and credit entries you would have two offsetting entries. Each offsetting the sum of the debits with a credit offset and the sum of the credit with a debit offset.

wadearnold:
> `POST /files/:id/balance` (We could offer parameters here if needed.)

`balance` or `balanced` ? 

Why is it a POST rather than a GET?

Will content type in the header be used to switch response types? `"Content-Type: text/plain"` to  `"Content-Type: application/json"` Or should we have `/files/:id/balanced/contents` as we do for files? 

> `POST /files/:id/split` (Assumes split into two files, one debit only and one credit only)

POST over GET? 

Should this create two new File ID's that represent the new resources that were created? We could return an array of files like `/files/create` returns but with new resources as well.

bkmoovio:
Are files being returned or json or both?

wadearnold:
> Are files being returned or json or both?

Both will need to be supported.

adamdecaf:
Credit and/or Debit offsets will only be added if they are needed. We wouldn't add zero-amount EntryDetails records. Sounds good. 

---

We can change marshaling of files based on content-type. I'm not quite sure how best to return multiple `text/plain` files though.. 

---

My thinking is that "balanced" ACH files is that they're just a presentational difference and nothing about their structure changes, so why do we need to keep track of what's supposed to be balanced? Why persist those modified files? We can do the compute to transform a non-balanced file into balanced easily. 

`GET` requests can be retried by load balancers or clients without a programmers knowledge, so they shouldn't be modifying the file. Thinking it over `GET` makes more sense as long as we don't modify the ACH files to balance or split. `POST` would make sense if we create new ACH files and store them. 

- `GET /files/:id/balanced`
- `GET /files/:id/split` 
   - This would return two new ACH files, but wouldn't store them (query param to store?)

bkmoovio:
Is an audit trail needed or is the offset record enough to identify that its changed?

adamdecaf:
The ACH service/library isn't the right place to store audit requirements. There's no persistence in this project. Paygate would be a better spot for that.

bkmoovio:
I was referring to balanced returning both the unbalanced file and balanced file.   Same with split, but your saying that those would be persisted somewhere else which is fine.

adamdecaf:
Not quite. I'm saying these balanced and split endpoints return files that aren't persisted anywhere by ACH. The original file isn't modified by these endpoints. The returned file is modified from the original, but not persisted. 

If a caller wants the original file, call `GET /files/:id` and if they want the file split or balanced file call the respective endpoints. This allows us to transform ACH files in multiple ways without modifying the original.

adamdecaf:
I could see supporting `PUT /files/:id/split` to create two new File objects in ACH (persisted in memory, like everything else).

adamdecaf:
@wadearnold and @bkmoovio I've updated this with what I think is a final version. Thoughts?

bkmoovio:
Setting up the EntryDetail and then setting to nil seems the best way to go if Amount is zero and then using the nil check to add the EntryDetail in the batch.  Do we have a test ach file for this?   For example starting with a file with PPD batches that are unbalanced resulting in a file with balanced PPD batches that can be read by the library? I would be nice to have to see the FileControl is accurate.  (It can be a separate issue if need be). We can add an example as part of Create examples directory #545 as well.

adamdecaf:
> Do we have a test ach file for this? For example starting with a file with PPD batches that are unbalanced resulting in a file with balanced PPD batches that can be read by the library? 

The [`TestBatch__CalculateBalancedOffsetDebitAndCredit` test](https://github.com/moov-io/ach/pull/537/files#diff-4e8c4de7ffbfe877f645136ab5bb1264R1200) and others account for this. Not from a `Reader` level, but the structs. 

> I would be nice to have to see the FileControl is accurate. (It can be a separate issue if need be). We can add an example as part of Create examples directory #545 as well.

The tests call `Validate()` and check FileControl's, so I think we're okay. There's a test that writes and replaces the offset record also.

adamdecaf:
I didn't think about `Reader` filling in the offset record. I suppose we could do that based on inverting the logic of matching the offset record.

bkmoovio:
I think we are ok, but I also think we want to highlight this functionality by having an example that creates a file that has balanced batches that users can have access too.  Maybe that is already created and I missed it.  

Currently devs can create balanced batches, without the new functionality.  Showing full tests/examples of the after snap shot of this enhancement would be great.  A dev can abstract from the tests included but lets make it really easy.  If possible even better to show the before and after snap shot.  (i.e.  An example of how it used to be done, this is how it can be done now).  We should highlight that devs no longer have to keep track of the entry amounts (sum), etc.  This is really cool automation and if we can convey that in meaningful way that would be great. 

For example does a dev need to include the conditional checks that are in the tests when creating balanced batches? 

I wasn't proposing Reader filling in the offset record, I was wondering if there was a test or example of a file created using the new functionality, that could be read by Reader. 

I wouldn't suggest any of this as critical path.

adamdecaf:
Tagging https://github.com/moov-io/moov-io/issues/22 to add an example file or two.

wadearnold (approved):

PR #551, opened 2019-06-21
comment code cov

bkmoovio:
Added tests and it is failing code cov


adamdecaf:
replaced by https://github.com/moov-io/ach/pull/552

Issue #556, opened 2019-06-24
fileCreationDate and fileCreationTime need more enforcement / error handling


eric-peterson-billgo:
**What were you trying to do?**
These two fields are shown in this example :  https://github.com/moov-io/ach/blob/master/test/testdata/ppd-valid.json and also this format is allowed to be posted in with json.

**What did you expect to see?**
These fields should be of the format yyMMdd and HHmm in all examples, and ideally when posting json this should also be enforced, otherwise you get back a ACH file this is likely invalid.

**What did you see?**

**How can we reproduce the problem?**
Take the ppd json file above, and post it in, then grab the ACH file back.  Then try to post that ACH file back in as text and you will see the year is not right.

adamdecaf:
Those fields support ISO8601 timestamps to allow languages to easier interop with our API. We return whatever format was given to us to keep the language's parser/reader happy.

There do seem to be missing validations and tests, which we can fix.

bkmoovio:
The following is the original converter code, maybe it can be re-implemented or some variation of it.

// formatSimpleDate takes a time.Time and returns a string of YYMMDD
func (v *Converters) formatSimpleDate(t time.Time) string {
	return t.Format("060102")
}

// parseSimpleDate returns a time.Time when passed time as YYMMDD
func (v *Converters) parseSimpleDate(s string) time.Time {
	t, _ := time.Parse("060102", s)
	return t
}

// formatSimpleTime returns a string of HHMM when  passed a time.Time
func (v *Converters) formatSimpleTime(t time.Time) string {
	return t.Format("1504")
}

// parseSimpleTime returns a time.Time when passed a string of HHMM
func (v *Converters) parseSimpleTime(s string) time.Time {
	t, _ := time.Parse("1504", s)
	return t
}

adamdecaf:
I have a different implementation of the fix. We need to handle both formats (timestamps and 4/6 number values).

bkmoovio:
Thanks Adam!

adamdecaf:
@eric-peterson-billgo Can you try out https://github.com/moov-io/ach/pull/558? It should handle the input a lot better and consistently format from ISO 8601 to YYMMDD/HHmm

PR #558, opened 2019-06-24
files: properly format FileCreation dates or times from multiple input formats

adamdecaf:
When `FileCreationTime` and `FileCreationDate` values are empty we are going to render them as tomorrow's appropriate value. This decision came from Wade, fyi. 

When the format is an ISO 8601 timestamp (common for Javascript and other languages) we need to handle that and convert to the format required by NACHA for ACH files. 

Also, clarify `m` vs `M` in time/date formats. `m` should refer to minutes and `M` to months. 

Issue: https://github.com/moov-io/ach/issues/556

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
       - "/Users/travis/Library/Caches/go-build"
 env:
   matrix:
-  - GO111MODULE=auto
+    - GO111MODULE=on
 before_install:
   - go get -u github.com/client9/misspell/cmd/misspell
   - go get -u golang.org/x/lint/golint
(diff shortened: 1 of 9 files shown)


adamdecaf:
@wadearnold Do you have more of an explanation as to why we pick tomorrow's date?

bkmoovio (approved):

bkmoovio (approved):

wadearnold:
I messed this up. 

File Header: 
File Creation Date and File Creation Time should default to the current date and time.  

Batch Header: 
Company Descriptive Date is the date you choose to identify the transactions in YYMMDD format. This date may be printed on the receivers’ bank statement by the RDFI
Effective Entry Date is the date transactions are to be posted to the receivers’ account. You almost always want the transaction to post as soon as possible, so put tomorrow's date in YYMMDD format. Technically it is the next Federal reserve date.

PR #577, opened 2019-07-18
WIP: Segment files

### LINKED ISSUE #528: File Creation - Segment Debit/Credit

bkmoovio:
Closes #528


bkmoovio:
That was delayed.

wadearnold (approved):

file.go
```diff
@@ -652,3 +654,99 @@ func (f *File) createFileADV() error {
 
 	return nil
 }
+
+// SegmentFile takes a valid ACH File and returns 2 segmented ACH Files, one ACH File containing credit entries and one
+// ACH File containing debit entries.
+func (f *File) SegmentFile() (*File, *File, error) {
```
adamdecaf: Can we clarify in the docs which file is debits-only and which is credits-only?

file.go
```diff
@@ -652,3 +654,99 @@ func (f *File) createFileADV() error {
...
+	}
+
+	creditFile := NewFile()
+	debitFile := NewFile()
+
+	for _, batch := range f.Batches {
+		bh := batch.GetHeader()
+
+		switch bh.ServiceClassCode {
+		case 200:
```
adamdecaf: Should we use the named constants here instead?

file.go
```diff
@@ -652,3 +654,99 @@ func (f *File) createFileADV() error {
...
+			debitBatch.Create()
+			debitFile.AddBatch(debitBatch)
+		case 220:
+			creditFile.AddBatch(batch)
+		case 225:
+			debitFile.AddBatch(batch)
+		}
+	}
+
+	// Additional Sorting to be FI specific
```
adamdecaf, line 709: I think we should pass through an optional struct (`SegmentConfig`) to allow for passing parameters later on.

We can't later change `SegmentFile()` to add parameters as that would break the API.

file_test.go
```diff
@@ -1056,3 +1057,105 @@ func TestFile__RemoveBatch(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Issue reading file: %+v \n", err)
+	}
+
+	// ensure we have a validated file structure
+	if achFile.Validate(); err != nil {
+		t.Fatalf("Could not validate entire read file: %v", err)
+	}
+
+	creditFile, debitFile, err := achFile.SegmentFile()
```
adamdecaf: Should we validate the number of batches in each file and some details?

file.go
bkmoovio: Yes

file.go
bkmoovio: Of course

file_test.go
bkmoovio: Do you want to validate against the original File?

file.go
bkmoovio: We could use goFix.  Nah, probably not a good idea, What properties should we include now? or is it just an Empty Struct for the time being?

file_test.go
bkmoovio: We could probably do that in the service_test as well.  Similar to

// TestGetBatches return a list of batches for the supplied file.id
func TestGetBatches(t *testing.T) {
	s := mockServiceInMemory()
	batches := s.GetBatches("98765")
	if len(batches) != 1 {
		t.Errorf("expected %s received %v", "1", len(batches))
	}
}

Either I guess is fine.

file.go
adamdecaf: We can include an empty struct. The point is allowing for expandability in the future because we can't break the API later on.

bkmoovio: ok, Will do!

PR #598, opened 2019-08-17
Redo examples

bkmoovio:


examples/example_ackRead_test.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 The Moov Authors
```
adamdecaf, line 1: Does this filename need to be something like `ach_read_example.go`?

bkmoovio: Good question.  I'm not sure but I used the sort package as an example for the naming convention.

https://golang.org/src/sort/

e.g.

example_keys_test.go 
example_multi_test.go 
example_search_test.go 
example_test.go 
example_wrapper_test.go

I was also using this as a reference. It is from 2015, so it could be outdated, but it's the first result returned in my google search using the keywords: golang examples 

https://blog.golang.org/examples

adamdecaf: Okay. I'd follow the stdlib as that's kept updated on each release

bkmoovio: yea, I agree it makes sense.  I was also looking at the go/doc source code and I'm going to make some adjustments based on that as well...

adamdecaf (approved):

PR #603, opened 2019-08-27
WIP Flatten batches

bkmoovio:


file.go
```diff
@@ -894,3 +896,79 @@ func segmentFileBatchAddADVEntry(creditBatch Batcher, debitBatch Batcher, entry
 		debitBatch.AddADVEntry(entry)
 	}
 }
+
+// OptimizeFile optimizes a file by flattening batches with the same BatchHeader data into one batch.
+func (f *File) OptimizeFile() (*File, error) {
```
adamdecaf: Could we call this FlattenBatches? Optimize is probably referencing cost relative to the Fed's prices, yes?

file.go
```diff
@@ -894,3 +896,79 @@ func segmentFileBatchAddADVEntry(creditBatch Batcher, debitBatch Batcher, entry
...
+
+	// Slice of BatchHeaders
+	sbh := make([]string, 0)
+	for _, b := range f.Batches {
+		bh := b.GetHeader().String()
+		sbh = append(sbh, bh[:94])
+	}
+
+	// Remove duplicate BatchHeader entries
+	sbh = removeDuplicateBatchHeaders(sbh)
```
adamdecaf, line 929: If it's easier I've used a `map[string]interface{}` to accumulate unique values before. You can overwrite based on key.

adamdecaf (commented):
(Sorry if you weren't looking for a review yet)

file.go
bkmoovio: yeah I started with that and then decided to go with a slice

file.go
bkmoovio: sure

file.go
bkmoovio: I created code to use a map.   Both are out there to see the difference

wadearnold (approved):

PR #605, opened 2019-08-29
reader: set EntryDetail.Category to Return when Addenda99 is present

adamdecaf:
Ray noticed this wasn't set in the JSON output after reading, which makes sense as we don't properly set the category.

### DIFF
--- a/entryDetail_test.go
+++ b/entryDetail_test.go
@@ -5,6 +5,10 @@
 package ach
 
 import (
+	"bytes"
+	"encoding/json"
+	"os"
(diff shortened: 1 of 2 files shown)


bkmoovio (approved):

bkmoovio (commented):
We should set Category for Addenda98 as well, either with this issue or a subsequent one.  Also probably for Forward entries as well as I don't think Reader was designed to define Category, although it may not be needed for Forward as it is a default in EntryDetail.

PR #609, opened 2019-08-29
WIP Flatten batches endpoint

bkmoovio:
I made this  PUT as a new File is created, but can adjust based on feedback


server/files.go
```diff
@@ -407,3 +406,54 @@ func decodeSegmentFileRequest(_ context.Context, r *http.Request) (interface{},
...
 	}, nil
 }
+
+type flattenBatchesRequest struct {
+	fileID    string
+	requestID string
+}
+
+type flattenBatchesResponse struct {
+	flattenFileID string `json:"flattenFileID"`
```
adamdecaf: The OpenAPI states we are returning an `ach.File` object, so we should have that here instead of just the File's ID.

adamdecaf (commented):
This looks good so far!

server/files.go
```diff
@@ -407,3 +406,54 @@ func decodeSegmentFileRequest(_ context.Context, r *http.Request) (interface{},
...
 	}, nil
 }
+
+type flattenBatchesRequest struct {
+	fileID    string
+	requestID string
+}
+
+type flattenBatchesResponse struct {
+	id  string `json:"id"`
```
adamdecaf: Go will only marshal exported fields, so this needs to be `ID`.

adamdecaf (approved):

PR #619, opened 2019-09-06
Flatten batches  IAT and ADV

bkmoovio:


file.go
```diff
@@ -917,47 +917,81 @@ func (f *File) FlattenBatches() (*File, error) {
...
+			for i, ofBatch := range of.IATBatches {
+				if strings.EqualFold(fbh, ofBatch.GetHeader().String()[:87]) {
+					iatEntries := iatBatch.GetEntries()
+					for _, iatEntry := range iatEntries {
+						// reset TraceNumber
+						iatEntry.TraceNumber = ""
+						of.IATBatches[i].AddEntry(iatEntry)
+					}
+				}
+				_ = of.IATBatches[i].Create()
```
adamdecaf, line 994: Should we check this error? It likely won't error because all we change is the TraceNumber, right?

file.go
```diff
@@ -1006,3 +1040,25 @@ func flattenBatchHeaderParse(record string) *BatchHeader {
 	bh.ODFIIdentification = bh.parseStringField(record[79:87])
 	return bh
 }
+
+// flattenIATBatchHeaderParse parses a string of IAT Batch Header data into a IATBatchHeader
+func flattenIATBatchHeaderParse(record string) *IATBatchHeader {
```
adamdecaf: What's the difference between this and `func (iatBh *IATBatchHeader) Parse(record string) `? It looks like this method duplicates a lot of the logic in Parse.

file.go
bkmoovio: that's correct, and the original file is validated

file.go
bkmoovio: sIATBh is 87 in length.

adamdecaf: Could we share the index lookups in a function though? I don't want that duplicated.

adamdecaf (approved):

PR #625, opened 2019-09-22
file: validate after reading JSON representation of ach.File

adamdecaf:


wadearnold (approved):
Wow I didn't know that validate was called. That could have created some bad issues.

PR #630, opened 2019-09-24
build: release v1.2.0-rc2

adamdecaf:


wadearnold (approved):
Let's open an issue for documenting this behaviour either in the readme or docs.

wadearnold (approved):

PR #643, opened 2019-10-16
addenda98: add helpers to parse and write CorrectedData 

### LINKED ISSUE #641: Addenda98: parse correction data

NOC files contain `CorrectedData string` which is a semi-structured format of correction data to update the ACH account with. We need to parse this field so paygate can update the respective object automatically for the user. 

I think creating a struct called `CorrectedData` with various fields which is returned from a parser helper method would satisfy this feature in paygate and for other users. The fields would all be optional and only populated based on the correction code. 

```Go
func ParseCorrectionData(in string) (*CorrectedData, error)

type CorrectedData struct {
   AccountNumber string
   RoutingNumber string
   Name string
   TransactionCode int
   Identification string
}
```

For more details on each code and what fields it would populate: https://github.com/moov-io/paygate/blob/master/internal/filetransfer/correction.go#L85

adamdecaf:
These helpers should let users bypass the code-specific parsing and give a better interface. 

Fixes: https://github.com/moov-io/ach/issues/641

### DIFF
--- a/addenda98.go
+++ b/addenda98.go
@@ -18,6 +18,7 @@
 package ach
 
 import (
+	"strconv"
 	"strings"
 	"unicode/utf8"
 )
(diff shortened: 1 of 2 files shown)


wadearnold (approved):
I love this!

PR #660, opened 2020-01-22
Bad characters in JSON marshaling Ç -> C

wadearnold:
entryAddendaÇount has a Ç and not a C. Fails marshalling to some users.


internal/iso3166/iso3166.go
```diff
@@ -76,7 +76,7 @@ var countryCodes = map[string]bool{
 	"CI": true, // Côte d'Ivoire
 	"HR": true, // Croatia
 	"CU": true, // Cuba
-	"CW": true, // Curaçao
+	"CW": true, // CuraCao
```
bendavies: incorrect replacement

adamdecaf: These files are created by `make generate` from an external source. I'll update it.

adamdecaf (approved):

PR #665, opened 2020-01-30
file: add ValidateWith to override specific default validations

adamdecaf:
We're starting to see more vendor or FI specific overrides and should keep those separate from the default validator set.

### DIFF
--- a/file.go
+++ b/file.go
@@ -505,7 +505,30 @@ func (f *File) SetHeader(h FileHeader) *File {
 //
 // Validate will never modify the file.
 func (f *File) Validate() error {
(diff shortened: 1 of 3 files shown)


wadearnold (approved):
It's to bad that we have to add this into the library but I think its the right thing to do. Where do we document that ValidateOpts exists?

adamdecaf:
This isn't ideal, but I'd rather offer opt-in degraded validation for vendor bugs using something like this. Eroding the default validation seems even worse. 

Assuming this looks good I was planning on writing up a page on `docs.moov.io/ach/`.

PR #673, opened 2020-02-24
files: keep TraceNumber in Flatten and Segment operations

### LINKED ISSUE #672: files: keep EntryDetail TraceNumbers during segment

Calling `f.SegmentFile(nil)` will wipe out existing EntryDetail `TraceNumber` values. This means any ability to set those for debugging/auditing later is also lost. 

We should keep those TraceNumber values by default.

adamdecaf:
`TraceNumber` fields are used for debugging and auditing from other systems. We should not wipe these by default, as that prevents reconciliation (and thus debug/audit operations). 

Fixes: https://github.com/moov-io/ach/issues/672


wadearnold:
# [Codecov](https://codecov.io/gh/moov-io/ach/pull/673?src=pr&el=h1) Report
> Merging [#673](https://codecov.io/gh/moov-io/ach/pull/673?src=pr&el=desc) into [master](https://codecov.io/gh/moov-io/ach/commit/6b49327f5ef44454de84ebdad2597f0c7f493c9a&el=desc) will **increase** coverage by `0.55%`.
> The diff coverage is `n/a`.


```diff
@@            Coverage Diff             @@
##           master     #673      +/-   ##
==========================================
+ Coverage   92.64%   93.19%   +0.55%     
==========================================
  Files          65       65              
  Lines        7379     5728    -1651     
==========================================
- Hits         6836     5338    -1498     
+ Misses        386      233     -153     
  Partials      157      157              
```

ViniciusMiana:
On the batch.go arround line 369, we are overiding the traceNumber inside the build method

batch.go
```diff
@@ -274,6 +276,10 @@ func (batch *Batch) Validate() error {
 	return errors.New("use an implementation of batch or NewBatch")
 }
 
+func (batch *Batch) SetValidation(opts *ValidateOpts) {
```
adamdecaf, line 279: I don't know if we want to add these options onto `File` and `Batcher`, but it would let us bypass validation to handle arbitrary Origin/Identification values.

adamdecaf:
The line mentioned above is: https://github.com/moov-io/ach/blob/ec4b52dc7973c157ffb2c074da21bff81078576b/batch.go#L383-L385

PR #699, opened 2020-04-01
reader: morph lines to 94 characters if they end in spaces

adamdecaf:


reader.go
```diff
@@ -118,8 +118,20 @@ func (r *Reader) Read() (File, error) {
 			if err := r.processFixedWidthFile(&line); err != nil {
 				r.errors.Add(err)
 			}
+
 		case lineLength != RecordLength:
-			r.errors.Add(r.parseError(NewRecordWrongLengthErr(lineLength)))
+			if lineLength > RecordLength {
+				// take the first 94 characters
```
wadearnold: Should we only take these if all the characters are blank? We should fail if data exists beyond RecordLength

wadearnold: I have never seen the use case that I am describing. Maybe I am overthinking this and we just got with this until we see a use case that requires the additional check.

wadearnold (approved):
This is better than what was previously there regardless of my comments.

reader.go
adamdecaf: I agree we should only trim extra spaces. Additional non-space characters would likely indicate an error in one of the fields.

wadearnold (approved):

PR #713, opened 2020-04-09
Handling 0 dollar/0 entry files returned as daily health pings from FISERV

prasadmahendra:
1. Don't validate file control fields BatchCount, EntryAddendaCount, EntryHash if there are no entries
2. Small error handling bug fix
3. Updated UTs for (1)


reader.go
```diff
@@ -152,11 +152,17 @@ func (r *Reader) Read() (File, error) {
...
 				r.errors.Add(r.parseError(err))
 			}
+
+			lineLength = len(line) // len(line) may have changed
 			r.line = line
-			// parse the line
-			if err := r.parseLine(); err != nil {
+
+			if lineLength != RecordLength {
 				r.errors.Add(r.parseError(NewRecordWrongLengthErr(lineLength)))
```
prasadmahendra: Noticed this error being returned incorrectly during our testing as a result of lineLength not getting updated after padding.

fileControl.go
```diff
@@ -116,21 +116,27 @@ func (fc *FileControl) Validate() error {
...
 	if fc.BlockCount == 0 {
 		return fieldError("BlockCount", ErrConstructor, fc.BlockCountField())
 	}
-	if fc.EntryAddendaCount == 0 {
-		return fieldError("EntryAddendaCount", ErrConstructor, fc.EntryAddendaCountField())
-	}
-	if fc.EntryHash == 0 {
-		return fieldError("EntryHash", ErrConstructor, fc.EntryAddendaCountField())
+
+	if fc.TotalCreditEntryDollarAmountInFile != 0 || fc.TotalDebitEntryDollarAmountInFile != 0 {
```
prasadmahendra, line 125: Is this a dangerous assumption to make?

adamdecaf:
@prasadmahendra Do you have an example health ping file?

reader.go
adamdecaf: I wanted to return the original length when an underlying error occurred as we're unsure what the underlying error would be.  I'm thinking we remove this change from the PR for now.

fileControl.go
adamdecaf: This is a zero-entry file, so maybe we should have an option on `ValidateOpts` to accept these invalid files? I need to read over the NACHA guidelines again to double check what they say to do in zero entry files.

prasadmahendra: If this is standard operating procedure (i.e Fiserv is a major banking core provider?) then I am not sure the value in making these only validate if certain ValidateOpts are set.

adamdecaf:
I've included that test file as an example for us, thanks!

fileControl.go
wadearnold: It's not a "standard" it is a hack by a major provider.

wadearnold (approved):

PR #716, opened 2020-04-14
server: read ValidateOpts in POST

adamdecaf:
This came up where a lot of HTTP clients don't allow `GET` requests to have request bodies. Since it's not uniformly allowed we need to accept `POST` as well. 

I don't like that there are now two endpoints/calls to validate a file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@ ADDITIONS
 - batches: Add `LiftEffectiveEntryDate()` to offer parsed `time.Time` values of `EffectiveEntryDate`
 - cmd/server: add version handler to admin HTTP server
 - file: add BypassDestinationValidation to ValidateOpts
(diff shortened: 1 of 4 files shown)


openapi.yml
```diff
@@ -203,6 +203,37 @@ paths:
...
+      responses:
+        '200':
+          description: File validated successfully without errors.
+          content:
+            application/json:
+              schema:
+                $ref: 'https://raw.githubusercontent.com/moov-io/api/master/openapi-common.yaml#/components/schemas/Error'
+        '400':
+          description: Validation failed. Check response for errors
+    post:
```
alovak, line 236: Does server route [here](https://github.com/moov-io/ach/blob/master/server/routing.go) need to be updated as well for POST request?

wadearnold (approved):
I wish that we could keep that a `GET` but I see no other way.

PR #720, opened 2020-05-14
ImmediateOriginField update

hitchr:
@adamdecaf  @wadearnold 
This allows for the ImmediateOriginField to be the full 10 characters that the bank is asking for in cases when the bank is asking for something other than a valid routing number in bTTTTAAAC format.

ValidateOpts.BypassOriginValidation allows us to get through the validate routine but ImmediateOriginField() is still formatting the field.

Since the NACHA specification says "This field may also be mutually defined between the ODFI and Originator", we need to be able to specify the full 10 characters requested.

### DIFF
--- a/fileHeader.go
+++ b/fileHeader.go
@@ -302,6 +302,9 @@ func (fh *FileHeader) ImmediateOriginField() string {
 	if fh.ImmediateOrigin == "" {
 		return strings.Repeat(" ", 10)
 	}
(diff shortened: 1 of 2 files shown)


adamdecaf:
Ah yea @hitchr that's a good catch! Thanks for the PR.

adamdecaf (approved):

PR #729, opened 2020-06-03
entries: allow custom trace numbers

adamdecaf:
Allowing custom TraceNumbers allows extra features to be written using ACH files. This is commonly used for products created in the past few years. 

There's also a branch with https://github.com/moov-io/ach/pull/728 -- https://github.com/moov-io/ach/tree/custom-entry-fields

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -353,8 +353,10 @@ func (batch *Batch) verify() error {
 	if err := batch.isBatchEntryCount(); err != nil {
 		return err
 	}
-	if err := batch.isSequenceAscending(); err != nil {
-		return err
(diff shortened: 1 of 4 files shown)


wadearnold (approved):

waltrcox:
assuming trace numbers are a subset of the customer ID, therefore it wouldn't matter if another customer used the same trace number (I.e., 273976369000001)?
do we generate a separate NACHA file for each customer ID? just want to make sure we don't send a file to process with two different customers both using the same trace number.

adamdecaf:
We hopped on a call, so to summarize trace numbers are always unique in a built file from this library. We'll need to look into spreading out trace numbers over the 000001-999999 range better to not duplicate of a lot of "01" trace numbers for an ODFI.

PR #735, opened 2020-06-26
batch: error if our offset account type is unknown

adamdecaf:
This came up as a linter error with OffsetAccountType(0) being an unexpected type casting.

https://travis-ci.com/github/moov-io/ach/jobs/354042702

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -1092,6 +1092,11 @@ func (b *Batch) upsertOffsets() error {
 		}
 	}
 
+	// Make sure the offset account type is valid
+	if err := b.offset.AccountType.validate(); err != nil {
+		return err
+	}
+
 	// Create our debit offset EntryDetail
 	debitED := createOffsetEntryDetail(b.offset, b)
 	debitED.TraceNumber = strconv.Itoa(lastTraceNumber(b.Entries) + 1)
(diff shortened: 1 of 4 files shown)


batch.go
```diff
@@ -1101,6 +1101,8 @@ func (b *Batch) upsertOffsets() error {
 		debitED.TransactionCode = CheckingDebit
 	case OffsetSavings:
 		debitED.TransactionCode = SavingsDebit
+	default:
+		return fmt.Errorf("unknown offset account type: %s", b.offset.AccountType)
```
wadearnold: Should these be `var` and declared somewhere else?

batch.go
```diff
@@ -1115,6 +1117,8 @@ func (b *Batch) upsertOffsets() error {
 		creditED.TransactionCode = CheckingCredit
 	case OffsetSavings:
 		creditED.TransactionCode = SavingsCredit
+	default:
+		return fmt.Errorf("unknown offset account type: %s", b.offset.AccountType)
```
wadearnold: `var` ?

wadearnold (approved):

batch.go
adamdecaf: A `var` so we can compare against it a bit better? We can do that.

PR #774, opened 2020-08-27
build: copy timezone data to final image

adamdecaf:


adamdecaf:
The issue was forgetting to set `odfi.cutoffs.timezone` in PayGate with a customer's setup.

PR #777, opened 2020-08-30
Added custom line ending support to Writer

ianberryman:
@adamdecaf 

Provides backwards-compatible support for custom line endings (assuming NewWriter is used to instantiate Writers)


wadearnold (approved):
Need to update the release notes and readme for usage.

PR #804, opened 2020-11-16
fix: add validation options to ach reader and currentBatch

vxio:
Bug: POST `/files/create?bypassOrigin=true` was returning error `(PPD) ODFIIdentificationField XXXX in header does not match entry trace number YYYY` when using a custom trace number.

This PR ensures the validation options set in the query params are passed through to the underlying reader and currentBatch.


server/files.go
```diff
@@ -170,6 +149,30 @@ func decodeCreateFileRequest(_ context.Context, request *http.Request) (interfac
...
+		// Attempt parsing body as an ACH File
+		r = bytes.NewReader(bs)
+		achReader := ach.NewReader(r)
+		achReader.SetValidation(req.validateOpts)
+	
+		f, err := achReader.Read()
+		req.File = &f
+		req.parseError = err
+	}
+
```
vxio: This code is generally unchanged, except the line where we set the validation options in the `ach.NewReader`: `achReader.SetValidation(req.validateOpts)`

adamdecaf (approved):

PR #817, opened 2020-12-29
Fix ImmediateDestination formatting when BypassDestinationValidation is true

### LINKED ISSUE #814: file-format: the destination field is not properly formatted when the BypassDestination validation option is set
ACH Version: `v1.5.2`

When setting `ach.ValidateOpts{ BypassDestinationValidation: true}`, the Destination field is not properly formatted. A `0` is set between the Origin and Destination fields. We should replace the `0` with a space.

We can reuse the same logic from `ImmediateOriginField()`:
https://github.com/moov-io/ach/blob/master/fileHeader.go#L301-L309

vxio:
Fixes #814 and added tests

### DIFF
--- a/fileHeader_test.go
+++ b/fileHeader_test.go
@@ -149,7 +149,7 @@ func TestFileHeader__ImmediateDestination(t *testing.T) {
 	}
 }
 
-func Test__trimRoutingNumberLeadingZero(t *testing.T) {
+func TestFileHeader__trimRoutingNumberLeadingZero(t *testing.T) {
 	tests := []struct {
 		input string
 		want  string
--- a/fileHeader_test.go
+++ b/fileHeader_test.go
@@ -81,11 +81,11 @@ func TestFileHeader__ImmediateOrigin(t *testing.T) {
 	if v := header.ImmediateOriginField(); v != " 123456789" {
 		t.Errorf("got %q", v)
 	}
-	header.ImmediateOrigin = trimImmediateOriginLeadingZero("0123456789") // 0 + routing number
+	header.ImmediateOrigin = trimRoutingNumberLeadingZero("0123456789") // 0 + routing number
 	if v := header.ImmediateOriginField(); v != " 123456789" {
 		t.Errorf("got %q", v)
 	}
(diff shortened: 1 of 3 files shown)


fileHeader_test.go
```diff
@@ -96,17 +96,81 @@ func TestFileHeader__ImmediateOrigin(t *testing.T) {
...
-	if v := trimImmediateOriginLeadingZero("0123456789"); v != "123456789" {
-		t.Errorf("got %q", v)
-	}
-	if v := trimImmediateOriginLeadingZero("0012345678"); v != "012345678" {
-		t.Errorf("got %q", v)
-	}
-	zeros := strings.Repeat("0", 10)
-	if v := trimImmediateOriginLeadingZero(zeros); v != zeros {
-		t.Errorf("got %q", v)
+func TestFileHeader__ImmediateDestination(t *testing.T) {
```
vxio, line 101: These tests are almost the same as the test above for `ImmediateOrigin` but we're using table-driven tests here for clarity; let me know if I should spin up another issue to convert the tests to match.

fileHeader_test.go
```diff
@@ -96,17 +96,81 @@ func TestFileHeader__ImmediateOrigin(t *testing.T) {
...
+			}
+
+			if v := header.ImmediateDestinationField(); v != tt.expected {
+				t.Errorf("want \"%v\", got \"%v\"", tt.expected, v)
+			}
+		})
+	}
+}
+
+func Test__trimRoutingNumberLeadingZero(t *testing.T) {
```
vxio: These checks used to live in `TestFileHeader__ImmediateOrigin` but I moved them here since the helper function is used for both origin and destination now.

adamdecaf (dismissed):

fileHeader_test.go
adamdecaf: This nice cleanup, thanks!

fileHeader_test.go
adamdecaf: Maybe call it `TestFileHeader__trimRoutingNumberLeadingZero(t *testing.T)` ?

adamdecaf (approved):

PR #820, opened 2021-01-04
errors: display parsing error and report error when extra characters are found

vxio:
Improvements:
- webui: display the error when parsing an ach file
- reader: return an error when parsing a one line file and the line length is not equal to `94`.


adamdecaf (commented):
Do you have a screenshot of this in webui?

vxio:
> Do you have a screenshot of this in webui?

image

adamdecaf (dismissed):

reader.go
```diff
@@ -138,11 +138,14 @@ func (r *Reader) Read() (File, error) {
 		lineLength := len(line)
 
 		switch {
-		case r.lineNum == 1 && lineLength > RecordLength && lineLength%RecordLength == 0:
-			if err := r.processFixedWidthFile(&line); err != nil {
+		case r.lineNum == 1 && lineLength > RecordLength:
+			extraChars := lineLength % RecordLength
+			if extraChars != 0 {
+				r.errors.Add(r.parseError(NewRecordWrongLengthErr(lineLength)))
+				r.errors.Add(fmt.Errorf("%d extra characters in ACH file", extraChars))
```
adamdecaf: Can we add a test verifying this error message?

vxio: sure thing

reader.go
```diff
@@ -138,11 +138,19 @@ func (r *Reader) Read() (File, error) {
 		lineLength := len(line)
 
 		switch {
-		case r.lineNum == 1 && lineLength > RecordLength && lineLength%RecordLength == 0:
-			if err := r.processFixedWidthFile(&line); err != nil {
+		case r.lineNum == 1 && lineLength > RecordLength:
+			extraChars := lineLength % RecordLength
+			if extraChars != 0 {
+				err := fmt.Errorf(
```
vxio, line 144: @adamdecaf I've replaced `RecordWrongLengthErr` with this error since the former will print out an incorrect message for this case, it would output something like `must be 94 characters and found 471`.

This change outputs `1 extra character(s) in ACH file: must be 470 but found 471`

adamdecaf: Cool!

adamdecaf (dismissed):

vxio:
@adamdecaf your approvals get auto-dismissed on new commits, I'll re-request a review

adamdecaf (approved):

PR #823, opened 2021-01-08
fix: make addenda optional on CIE batches

Deiz:
This aligns with the spec, which stipulates that while CIE batches can
only have a maximum of one addenda record, including addenda is
optional.


adamdecaf (approved):

vxio (approved):

vxio:
Appreciate the PR!

PR #824, opened 2021-01-08
fix: Add ` to the alphanumeric character set

Deiz:
Test files have been updated to use ASCII DEL as an example invalid
character.

The spec has this to say re: the alphanumeric character set:

0-9, A-Z, a-z, space, EBCDIC values greater than hexadecimal "3F", ASCII
values greater than hexadecimal "1F"


test/testdata/Iat-invalidAddenda13.ach
```diff
@@ -4,7 +4,7 @@
 710ANN000000000000100000928383-23938          BEK Enterprises                          0000001
 711BEK Solutions                      15 West Place Street                             0000001
 712JacobsTown*PA\                     US*19305\                                        0000001
-713Wells`Fargo                        01231380104                         US           0000001
+713WellsFargo                        01231380104                         US           0000001
```
vxio: This is a bit tricky, but we should avoid using the `DEL` character as an invalid char in this testdata. Is there another invalid character that would be more explicit?

validators.go
```diff
@@ -27,8 +27,8 @@ import (
 )
 
 var (
-	upperAlphanumericRegex = regexp.MustCompile(`[^ A-Z0-9!"#$%&'()*+,-.\\/:;<>=?@\[\]^_{}|~]+`)
-	alphanumericRegex      = regexp.MustCompile(`[^ \w!"#$%&'()*+,-.\\/:;<>=?@\[\]^_{}|~]+`)
+	upperAlphanumericRegex = regexp.MustCompile(`[^ A-Z0-9!"#$%&'()*+,-.\\/:;<>=?@\[\]^_{}|~` + "`" + `]+`)
```
vxio: Using `fmt.Sprintf` should work here too:
```go
	upperAlphanumericRegex = regexp.MustCompile(fmt.Sprintf(
		`[^ A-Z0-9!"#$%%&'()*+,-.\\/:;<>=?@\[\]^_{}|~%s]+`, "`",
	))

```

test/testdata/Iat-invalidAddenda13.ach
Deiz: Any of the 0x00 -> 0x1F control characters would suffice, too.

Unfortunately with this change, we have no more invalid printable ASCII characters.

validators.go
Deiz: Seems a little more readable, I'll adopt this.

test/testdata/Iat-invalidAddenda13.ach
Deiz: One other option would be to use an "extended ASCII" character and re-encode the test files as ISO-8559 - we could then use any of the characters in e.g. the ISO 8859-1 code page, which Go will still parse as a single byte rune.

vxio: Could we try the bell character? `BEL = 0x07 // '\a' Bell`
It should show up in Github as
```bash

```

validators_test.go
```diff
@@ -119,3 +119,42 @@ func TestValidators__validateSimpleTime(t *testing.T) {
...
+		if i < 0x20 || i > 0x7E {
+			if err == nil {
+				t.Errorf("expected rune %x to be non-alphanumeric", i)
+			}
+		} else {
+			if err != nil {
+				t.Errorf("expected rune %x to be alphanumeric: %v", i, err)
+			}
+		}
+	}
```
vxio: I think you can simplify this to:
```go
		err := v.isAlphanumeric(string(rune(i)))
		if (i < 0x20 || i > 0x7E) && err == nil {
			t.Errorf("expected rune %x to be non-alphanumeric", i)
		} else if err != nil {
			t.Errorf("expected rune %x to be alphanumeric: %v", i, err)
		}

```

validators_test.go
```diff
@@ -119,3 +119,42 @@ func TestValidators__validateSimpleTime(t *testing.T) {
...
+}
+
+// Ensure that ASCII characters from 0x20 to 0x60 and 0x7B to 0x7E are considered upper case alphanumeric.
+func TestValidators__isUpperAlphanumeric(t *testing.T) {
+	v := validator{}
+
+	for i := 0; i < 255; i++ {
+		chr := string(rune(i))
+		err := v.isUpperAlphanumeric(chr)
```
vxio: Same comment above about the if-else logic

test/testdata/Iat-invalidAddenda13.ach
Deiz: If you'd like, sure. However, it'll make your shell bleep at you when you look at `git log -p` :stuck_out_tongue:

validators_test.go
Deiz: Not quite, because that will break if i is outside the 0x20 - 0x7E range - in that scenario, we expect an error. But I can still simplify a little.

test/testdata/Iat-invalidAddenda13.ach
adamdecaf: If we can let's avoid bell/alarm characters. Those have caused annoyances in the past on other projects. Side effects such as triggering hardware sounds isn't ideal.

vxio: lol I didn't know that.

Also, I made this PR: https://github.com/moov-io/ach/pull/827

Deiz: Yeah, that's simpler and won't trip up editors. As long as we're good with the files being UTF-8, happy to throw a multi-byte character in instead of an ASCII control character.

vxio: I tested it locally, but unfortunately, `Parse()` still parses the record as bytes. So using an invalid character UTF-8 character in tests won't work for now.
Link to the Parse() method: https://github.com/moov-io/ach/blob/080e8fcf72fda916f380335bc5b38b63259ef4d7/fileHeader.go#L123

Instead of the `bell` to avoid the sound, could we try `ESC = 0x1B // '\e' Escape`? I tested it locally, and it works with the test:

image

@Deiz thanks for being patient while we try out these weird edge cases :)

Deiz: ESC may still trigger terminal misbehaviour when printed, so I went with SUB (the replacement character, �) which a) is printable, b) is disallowed by the spec because it's below 0x1F in ASCII.

validators_test.go
```diff
@@ -119,3 +119,38 @@ func TestValidators__validateSimpleTime(t *testing.T) {
 		}
 	}
 }
+
+// Ensure that ASCII characters from 0x20 to 0x7E are considered alphanumeric.
+func TestValidators__isAlphanumeric(t *testing.T) {
```
vxio, line 123: Nice refactor. Also, I would recommend combining these two tests into one using table-driven tests since they share the same test logic, here's an example:

```go
func TestValidators__isAlphanumeric(t *testing.T) {
	v := validator{}

	tests := []struct {
		name      string
		checkFunc func(string) error
		shouldErr func(i int) bool
	}{
		// Ensure that ASCII characters from 0x20 to 0x7E are considered alphanumeric.
		{
			name:      "is alphanumeric",
			checkFunc: v.isAlphanumeric,
			shouldErr: func(i int) bool {
				return i < 0x20 || i > 0x7E
			},
		},
		// Ensure that ASCII characters from 0x20 to 0x60 and 0x7B to 0x7E are considered upper case alphanumeric.
		{
			name:      "is upper alphanumeric",
			checkFunc: v.isUpperAlphanumeric,
			shouldErr: func(i int) bool {
				return i < 0x20 || i > 0x7E || (i > 0x60 && i < 0x7B)
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			for i := 0; i < 255; i++ {
				chr := string(rune(i))
				err := tt.checkFunc(chr)
				shouldError := tt.shouldErr(i)

				if shouldError && err == nil {
					t.Errorf("expected rune %x (%s) to be non-alphanumeric", i, chr)
				} else if !shouldError && err != nil {
					t.Errorf("expected rune %x (%s) to be alphanumeric: %v", i, chr, err)
				}
			}
		})
	}
}
```

Deiz: Adopted this.

vxio (approved):
Nice work!

PR #826, opened 2021-01-08
fix: allow POS batches to contain credits

Deiz:
This also loosens the service class code and permits POS batches to be
mixed, credit-only, or debit-only.

While the spec initially defines POS entries as debits, it when
elaborates:

Also an adjusting or other credit Entry related to such debit Entry,
transfer of funds, or obligation.


adamdecaf (approved):

vxio (approved):
👍🏻

vxio:
Thanks again for the PR!

PR #844, opened 2021-01-29
Respect batch numbers provided when creating file

### LINKED ISSUE #841: Ability to designate starting point of Batch Sequence Number
Version: v1.6.0

What were you trying to do?
Designate the batchnumber or starting point of the batch sequence via the create file api

What did you expect to see?
The batch numbers passed in with the batch headers to be used

What did you see?
Batch numbers start at 1 and iterate

This is a real world requirement I have where my downstream consumers use the batch number as a dedup mechanism  for the day so I need to be able to define the starting point for the batch

Dartths:
Uses batch numbers provided by the user in batch header record rather than overriding them with a sequence starting at 1.
Performs a validation on the batch numbers to ensure they are sequential.

Fixes #841


file_test.go
```diff
@@ -1965,3 +1965,23 @@ func TestFile__FileHeaderFormattingWithValidation(t *testing.T) {
 		t.Fatalf("want \"%v\", got \"%v\"", want, s[:len(want)])
 	}
 }
+
+// testFileEntryHash validates entry hash
+func TestFileiSequenceAscending(t *testing.T) {
```
vxio:
```suggestion
func TestFile__AscendingBatchSequence(t *testing.T) {
```

merge_test.go
```diff
@@ -265,3 +263,23 @@ func TestMergeFiles__invalid(t *testing.T) {
 		t.Errorf("expected error: len(out)=%d error=%v", len(out), err)
 	}
 }
+
+func populateFileWithMockBatches(noOfBatches int, file *File) {
```
vxio:
```suggestion
func populateFileWithMockBatches(numBatches int, file *File) {
```

merge_test.go
```diff
@@ -265,3 +263,23 @@ func TestMergeFiles__invalid(t *testing.T) {
...
+		header.ServiceClassCode = 225
+		header.CompanyName = "Example Company"
+		header.CompanyIdentification = "132465"
+		header.CompanyEntryDescription = "Example Description"
+		header.ODFIIdentification = "12104288"
+		batch, _ := NewBatch(header)
+		batch.AddEntry(entryDetail)
+		batch.GetHeader().BatchNumber = i
+		batch.GetControl().BatchNumber = i
+		batch.Create()
```
vxio: We should check the error here too.

merge_test.go
```diff
@@ -265,3 +263,23 @@ func TestMergeFiles__invalid(t *testing.T) {
 		t.Errorf("expected error: len(out)=%d error=%v", len(out), err)
 	}
 }
+
+func populateFileWithMockBatches(noOfBatches int, file *File) {
+	var startSeq = file.Batches[0].GetHeader().BatchNumber + 1
```
vxio: Shouldn't the `startSeq` start from the file's last batch? In turn, you should be able to use:
```go
	lastBatchIdx := len(file.Batches) - 1
	var startSeq = file.Batches[lastBatchIdx].GetHeader().BatchNumber + 1
```

merge_test.go
```diff
@@ -265,3 +263,23 @@ func TestMergeFiles__invalid(t *testing.T) {
...
+	var entryDetail = file.Batches[0].GetEntries()[0]
+	for i := startSeq; i < (noOfBatches + startSeq); i++ {
+		header := mockBatchHeader()
+		header.StandardEntryClassCode = "PPD"
+		header.ServiceClassCode = 225
+		header.CompanyName = "Example Company"
+		header.CompanyIdentification = "132465"
+		header.CompanyEntryDescription = "Example Description"
+		header.ODFIIdentification = "12104288"
+		batch, _ := NewBatch(header)
```
vxio: We should check the error here; we can do that by passing in `t *testing.T` to this method.

fileErrors.go
```diff
@@ -122,3 +122,23 @@ func NewErrFileCalculatedControlEquality(field string, calculated, control int)
...
+
+// ErrFileBatchNumberAscending is the error given when the batch numbers in a file are not in ascending order
+type ErrFileBatchNumberAscending struct {
+	Message       string
+	PreviousBatch int
+	CurrentBatch  int
+}
+
+// NewErrFileBatchNumberAscending creates a new error of the ErrFileBatchNumberAscending type
+func NewErrFileBatchNumberAscending(field string, previous, current int) ErrFileBatchNumberAscending {
```
vxio: I don't think we need to pass in `field string` here since it'll always be called `BatchNumber`.

I recommend using the message:
```go
		Message:       fmt.Sprintf("Batch numbers must be in ascending order, batch %v is less than or equal to the previous batch: %v", current, previous),
```

file.go
```diff
@@ -1125,6 +1134,20 @@ func (f *File) FlattenBatches() (*File, error) {
...
+
+	lastSeq := 0
+	for _, batch := range f.Batches {
+		current := batch.GetHeader().BatchNumber
+		if current <= lastSeq {
+			return NewErrFileBatchNumberAscending("BatchNumber", lastSeq, current)
+		}
+		lastSeq = current
+	}
+	return nil
```
vxio: nit: some minor formatting changes 
```suggestion
	lastSeq := 0
	for _, batch := range f.Batches {
		current := batch.GetHeader().BatchNumber
		if current <= lastSeq {
			return NewErrFileBatchNumberAscending("BatchNumber", lastSeq, current)
		}
		
		lastSeq = current
	}
	
	return nil
```

file_test.go
vxio: I recommend updating this to use table-driven tests so we can test various success and error cases, here's an example of what that'd look like:
```go
func TestFileiSequenceAscending(t *testing.T) {
	tests := []struct {
		desc        string
		sequence    []int
		expectedErr error
	}{
		{
			desc:        "success: strictly ascending",
			sequence:    []int{2, 5, 7, 9},
			expectedErr: nil,
		},
		{
			desc:        "descending then ascending",
			sequence:    []int{3, 2, 4},
			expectedErr: NewErrFileBatchNumberAscending("BatchNumber", 3, 2),
		},
		{
			desc:        "ascending then descending",
			sequence:    []int{2, 3, 2},
			expectedErr: NewErrFileBatchNumberAscending("BatchNumber", 3, 2),
		},
		{
			desc:        "equal values",
			sequence:    []int{2, 3, 3, 4},
			expectedErr: NewErrFileBatchNumberAscending("BatchNumber", 3, 3),
		},
	}

	for _, tt := range tests {
		t.Run(tt.desc, func(t *testing.T) {
			file := mockFilePPD()

			for _, num := range tt.sequence {
				mockBatch := mockBatchPPD()
				mockBatch.GetHeader().BatchNumber = num
				mockBatch.GetControl().BatchNumber = num
				//mockBatch.Create()
				file.AddBatch(mockBatch)
			}
			if err := file.Create(); err != nil {
				t.Fatal(err)
			}

			err := file.Validate()
			if err != tt.expectedErr {
				t.Errorf("got: %v, want: %v", err, tt.expectedErr)
			}
		})
	}
}

```

vxio (commented):
Looking good! I have a few comments.

file_test.go
Dartths: Done

merge_test.go
Dartths: Done

merge_test.go
Dartths: Done

fileErrors.go
Dartths: Good idea, done.

merge_test.go
Dartths: I'm not really sure what you mean to check here? Could you explain a bit more?

vxio: since this is a test helper, you can change the method signature to 
```go
func populateFileWithMockBatches(t *testing.T, numBatches int, file *File)
...
batch, err := NewBatch(header)
if err != nil {
  t.Fatal(err)
}
...
```

vxio (approved):
Great job -- approved!

vxio:
@Dartths I think you accidentally closed the PR?

Reopening it; after the tests pass, I'll merge it in.

PR #846, opened 2021-02-02
build: set GITHUB_TOKEN in make check

adamdecaf:


vxio (approved):

adamdecaf:
I think https://github.com/moov-io/infra/pull/159 is a better fix for all our repositories. That check isn't very helpful as we are aware of new releases.

Issue #829, opened 2021-01-11
JSON Specification?


5outh:
ACH Version: `v1.5.2`

**What were you trying to do?**

Read from/write to ACH files using a Haskell data type with  `achcli`'s JSON format as an intermediate representation.

In order to do this effectively, I need to know exactly what the data type looks like. I ran the parser on two example files (one normal transaction, one return) and I can see that they differ - the return JSON contains fields that the normal transaction's JSON format does not (for example `addenda99`). How can I get more detail on how exactly this JSON is formatted without digging into the source code?

**What did you expect to see?**

A specification of the JSON format for `achcli` (data types, required, nullable, or optional)

**What did you see?**

I do not see a formal specification.

**How can we reproduce the problem?**

N/A

vxio:
Hey @5outh, we're working on updating our [open api spec](./openapi.yaml) to specify the data types on each of our models. That way, you can look there to see a field's type and whether it's required/nullable/optional. Would that satisfy your use case?

adamdecaf:
If there are specific examples where the OpenAPI spec is wrong please let us know.

PR #848, opened 2021-02-02
update openapi spec to add details to data types

vxio:
Helps #829

### DIFF
--- a/file_test.go
+++ b/file_test.go
@@ -424,9 +424,34 @@ func BenchmarkFileReturnEntries(b *testing.B) {
 // vince tester todo: delete
 func TestX(t *testing.T) {
 	f := NewFile()
-	b, _ := json.MarshalIndent(&f, "", " ")
-	fmt.Printf("%s", b)
+	if err := f.Create(); err != nil {
+		fmt.Println(err)
+	}
+	if err := f.Validate(); err != nil {
+		fmt.Println(err)
+	}
 
+	b, err := json.MarshalIndent(&f, "", " ")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Printf("new file:\n %s", b)
+
+	path := filepath.Join("test", "testdata", "ppd-valid.json")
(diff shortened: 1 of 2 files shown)


file_test.go
```diff
@@ -421,6 +421,14 @@ func BenchmarkFileReturnEntries(b *testing.B) {
 	}
 }
 
+// vince tester todo: delete
+func TestX(t *testing.T) {
```
vxio: Ignore this: this is just my playground for inspecting the JSON form of a file

openapi.yaml
```diff
@@ -596,11 +604,14 @@ components:
           type: string
           description: Incremented value for each file for RDFI's.
           example: "0"
+          nullable: true
       required:
         - immediateOrigin
         - immediateOriginName
         - immediateDestination
         - immediateDestinationName
+        - fileCreationTime
```
vxio, line 612: @adamdecaf per the openapi spec: `Note that an empty string "" is a valid string unless minLength or pattern is specified.`

So would it make sense to add this field to `required` since we don't specify `omitempty` on it?

ref: https://swagger.io/docs/specification/data-models/data-types/

vxio: ^^ nvm, ignore the comment above, I'm going to add the `pattern` specifier to fields like this

adamdecaf: Do we set a value if that string is empty? I would think the library would populate it on `Create()` if empty.

vxio: For `FileCreationTime`, it'll just be `""`

adamdecaf: Oh it's because when we write the FileHeader an empty string is written as the current time. It's fine if that is left out of a request. We'll populate the correct value when constructing the file.

https://github.com/moov-io/ach/blob/master/fileHeader.go#L339

vxio: oh i see, thanks for the link!

adamdecaf (approved):

adamdecaf:
Thanks for grabbing this!

PR #856, opened 2021-02-15
optimized and refactored FlattenBatches(); added benchmarks

### LINKED ISSUE #854: optimize: File.FlattenBatches() has exponential time complexity based on # of batches in an ACH file
Based on benchmarks, it looks like `File.FlattenBatches()` has an **exponential time complexity** based on the # of batches in the file. We should be able to refactor the implementation to get **linear time complexity**.

---
**Benchmark Results**:

To create, flatten, and read contents using the HTTP server
```
Benchmarks for ACH file with 75000 transactions split up in 50 batches
goos: darwin
goarch: amd64
pkg: github.com/moov-io/ach/server
BenchmarkCreateBigFile__Library
BenchmarkCreateBigFile__Library-16    	       6	 180223169 ns/op
BenchmarkCreateBigFile__Server
BenchmarkCreateBigFile__Server-16     	       3	 383476791 ns/op
Benchmark__FlattenBigFile
Benchmark__FlattenBigFile-16          	       1	2571188081 ns/op
Benchmark__All
Benchmark__All-16                     	       1	2984088173 ns/op // ~3 seconds
PASS
--
Benchmarks for ACH file with 75000 transactions split up in 500 batches
goos: darwin
goarch: amd64
pkg: github.com/moov-io/ach/server
BenchmarkCreateBigFile__Library
BenchmarkCreateBigFile__Library-16    	       6	 172432146 ns/op
BenchmarkCreateBigFile__Server
BenchmarkCreateBigFile__Server-16     	       3	 389614095 ns/op
Benchmark__FlattenBigFile
Benchmark__FlattenBigFile-16          	       1	21308247045 ns/op
Benchmark__All
Benchmark__All-16                     	       1	22460986042 ns/op // ~22 seconds
PASS
```

Source code is here: https://github.com/moov-io/ach/blob/vince-benchmark/server/benchmark_test.go

vxio:
Fixes #854 

This PR brings down the benchmark for `FlattenFiles()` from `~22s` to `~.3s`, a 100x improvement:
```
Benchmarks for ACH file with 75,000 transactions split up in 1,000 batches
goos: darwin
goarch: amd64
pkg: github.com/moov-io/ach/server
Benchmark__FlattenBigFile
Benchmark__FlattenBigFile-16    	       3	 334161791 ns/op
```

### DIFF
--- a/file.go
+++ b/file.go
@@ -1081,14 +1081,14 @@ func (f *File) FlattenBatches() (*File, error) {
 		}
 	}
 
-	IATBatchesByHeader := make(map[string]IATBatch)
+	IATBatchesByHeader := make(map[string]*IATBatch)
 	if f.IATBatches != nil {
 		for _, b := range f.IATBatches {
 			bhKey := getHeader(b.GetHeader())
 			_, found := IATBatchesByHeader[bhKey]
 			if !found {
 				newBatch := NewIATBatch(b.GetHeader())
-				IATBatchesByHeader[bhKey] = newBatch
+				IATBatchesByHeader[bhKey] = &newBatch
 				out.AddIATBatch(newBatch)
 			}
 
@@ -1117,7 +1117,7 @@ func (f *File) FlattenBatches() (*File, error) {
 		go func(b IATBatch) {
 			defer wg.Done()
 			_ = b.Create()
-		}(b)
+		}(*b)
 	}
 
 	wg.Wait()
--- a/file.go
+++ b/file.go
@@ -1103,12 +1103,16 @@ func (f *File) FlattenBatches() (*File, error) {
 
 	// Set batches to valid state
 	wg := sync.WaitGroup{}
+	doneCh := make(chan struct{}, 1)
+	errCh := make(chan error, 1)
 
 	wg.Add(len(batchesByHeader))
 	for _, b := range batchesByHeader {
 		go func(b Batcher) {
 			defer wg.Done()
-			_ = b.Create()
+			if err := b.Create(); err != nil {
+				errCh <- err
+			}
 		}(b)
 	}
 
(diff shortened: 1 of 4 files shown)


adamdecaf (dismissed):

file.go
```diff
@@ -1103,24 +1103,41 @@ func (f *File) FlattenBatches() (*File, error) {
...
+		wg.Wait()
+		close(doneCh)
+	}()
+
+	// Block til we're done creating batches or an error occurs
+	select {
+	case <-doneCh:
+		break
+	case err := <-errCh:
+		return nil, fmt.Errorf("creating batch: %w", err)
```
adamdecaf: If we were discarding errors before wouldn't this break the ABI (application behavior) of `FlattenBatches()`? Errors that were previously swallowed are now reported which could break calling code.  

I agree we should be checking these, but I can see why we skipped errors before. It was easier to "toss a bunch of batches into a file" and fix them after flattening.

adamdecaf: I almost want to suggest we add a `FlattenBatchesWith(opts *FlattenBatchesOpts)` and have `FlattenBatches()` supply `nil`. 

`FlattenBatchesOpts` can have an `ExposeErrors bool` and in the future stuff like `MaxBatches int` which has been requested before.

vxio: ohhh i like that idea

vxio: I'm going to implement what you suggested

adamdecaf: Sounds good. Can we restore the error discarding behavior and merge this? A performance improvement (without changing ABI) can go out as a bug fix.

FlattenBatchesWith would require a minor release.

vxio: Sure thing, I'll create an issue to track

adamdecaf (approved):

file.go
hampgoodwin: If this is a pattern you consider helpful or extending, I'm a huge fan of https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

PR #870, opened 2021-03-15
Override batch numbers in MergeFiles()

### LINKED ISSUE #867: file: MergeFiles can produce invalid files
ACH Version: `v1.6.2`

When calling `MergeFiles(...)` with multiple files you can get an `ach.File` out does not validate. This is from the ordering of input files that are appended to the output file, but done so in an incorrect ordering. 

`MergeFiles(..)` should product valid files.

**What did you see?**
`MergeFiles(..)` produced an file which was not valid.

vxio:
Fixes #867

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -71,9 +71,18 @@ func MergeFiles(files []*File) ([]*File, error) {
 		}
 	}
 
-	// TODO(adam): We should also look at consolidating EntryDetail records inside Batches
+	// Return LOC-maxed files and merged files
+	out := append(fs.locMaxed, fs.outfiles...)
 
-	return append(fs.locMaxed, fs.outfiles...), nil // return LOC-maxed files and merged files
+	// Override batch numbers to ensure they don't collide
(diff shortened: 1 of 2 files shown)


adamdecaf (dismissed):

merge.go
```diff
@@ -71,9 +71,18 @@ func MergeFiles(files []*File) ([]*File, error) {
 		}
 	}
 
-	// TODO(adam): We should also look at consolidating EntryDetail records inside Batches
+	// Return LOC-maxed files and merged files
+	out := append(fs.locMaxed, fs.outfiles...)
 
-	return append(fs.locMaxed, fs.outfiles...), nil // return LOC-maxed files and merged files
+	// Override batch numbers to ensure they don't collide
```
adamdecaf, line 78: Can we document this on the GoDoc?

vxio: Updated that!

adamdecaf (approved):

PR #885, opened 2021-03-28
Remove print statement from init() to avoid printing benchmark info on other tests

vxio:


vxio:
Just have to rebase and should fix the build issue.

adamdecaf (approved):

Issue #863, opened 2021-02-24
file: FlattenBatches produces incorrectly ordered EntryDetail records


adamdecaf:
ACH Version: `v1.6.2`

**What were you trying to do?**

**What did you expect to see?**

**What did you see?**

**How can we reproduce the problem?**

vxio:
I'll look into this, this may not be an issue anymore given the recent refactor to FlattenBatches: https://github.com/moov-io/ach/commit/bdc08d224d6ff16e9664a8ed51fb1959e578410c

Do you have an example file I could test against?

adamdecaf:
Fixed in https://github.com/moov-io/ach/pull/894

PR #894, opened 2021-04-13
test/issues: add failing test for flatten batches out-of-order

adamdecaf:
Issue: https://github.com/moov-io/ach/issues/863


adamdecaf:
It looks like https://github.com/moov-io/ach/commit/8a3eeb88efcb8cdf25fe7d3f09360c3ee4a544a6 was a partial fix for this issue, but we're still merging/flattening files with out-of-order trace numbers. I think we'll have to re-order those inside `FlattenBatches()`.

adamdecaf:
Actually, should we re-order the entries? That way we aren't modifying the file contents at all. Most people will use the `TraceNumber` as a unique value so our library changing that will break their use-cases.

adamdecaf:
```go
func sortEntries(entries []*EntryDetail) []*EntryDetail {
	sort.Slice(entries[:], func(i, j int) bool {
		return entries[i].TraceNumber < entries[j].TraceNumber
	})
	return entries
}
```

This lets us easily update the entry ordering but with `Batcher` it's not a straightforward usage. I think we can do a quick `b, ok := f.Batches[i].(*Batch); ok` to sort them with `b.Entries = sortEntries(..)`. I'll try that out. Each BatchXXX inherits `Batch` so it _should_ work.

adamdecaf:
@vxio can you give me your thoughts on this?

vxio:
> Actually, should we re-order the entries? That way we aren't modifying the file contents at all. Most people will use the `TraceNumber` as a unique value so our library changing that will break their use-cases.

I think it's fine if we re-order the entries. Similar to recent fixes we made to `MergeFiles(..)`, we have to do some reordering/overriding to ensure items aren't out-of-order or collide.

I like your idea of sorting the entries by traceNumber.

entryDetail.go
```diff
@@ -604,3 +605,10 @@ func (ed *EntryDetail) addendaCount() (n int) {
 	}
 	return n
 }
+
+func sortEntries(entries []*EntryDetail) []*EntryDetail {
```
vxio:
```suggestion
func sortEntriesByTraceNumber(entries []*EntryDetail) []*EntryDetail {
```

file.go
```diff
@@ -1084,6 +1084,20 @@ func (f *File) FlattenBatches() (*File, error) {
 			}
 		}
 	}
+	for i := range out.Batches {
+		bh := out.Batches[i].GetHeader()
+		if bh.StandardEntryClassCode != "ADV" {
+			batch, _ := NewBatch(bh)
+			entries := sortEntries(out.Batches[i].GetEntries())
```
vxio: We should update the GoDoc to reflect this too

adamdecaf: Fixed, thanks!

vxio (approved):
Looks good to me!

Should I create an issue to add ADV and IAT support in the near future?

adamdecaf:
Sure. I figure IAT would be requested first. I'm not sure how popular ADV would be for this function.

PR #895, opened 2021-04-14
merge: count individual records rather than write/render File

### LINKED ISSUE #862: merge: lineCount can be improved
ACH Version: `v1.6.2`

When calling `MergeFiles` there's a function called `lineCount` that is used for when ACH files reach their Nacha designated limits. This is used to comply with Nacha guidelines that limit file lengths to 10k lines.

The implementation of `lineCount` is a) inefficient and b) requires the file completes a call to `Create()` without error. 

**What did you see?**
`lineCount` should be as fast as possible. On large arrays of files the performance will not be optimal.

adamdecaf:
This comes about where trying to Write an invalid file fails so MergeFiles can often fail.

Fixes: https://github.com/moov-io/ach/issues/862

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -18,8 +18,6 @@
 package ach
 
 import (
-	"bufio"
-	"bytes"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


vxio (dismissed):
Nice!

adamdecaf:
Before:

```
$ go test -bench BenchmarkMergeFiles__lineCount
BenchmarkMergeFiles__lineCount-16    	      51	  41916935 ns/op
```

After:

```
$ go test -bench BenchmarkMergeFiles__lineCount 
BenchmarkMergeFiles__lineCount-16    	   44428	     26693 ns/op
```

vxio (approved):
Awesome benchmark results!!

PR #899, opened 2021-04-15
validation: add ability to bypass matching company ID in batch header and control

### LINKED ISSUE #898: validation: add an option to bypass validation that requires the Company ID in the batch header and control to match.
ACH Version: `1.6.5`

Create an ACH file where the Company ID in the Batch Header and Batch Control do not match.

The file to pass validation.

**What did you see?**
A validation error that requires the Company ID to match in the Batch Header and Control.

vxio:
Fixes #898

### DIFF
--- a/file.go
+++ b/file.go
@@ -564,6 +564,10 @@ type ValidateOpts struct {
 
 	// AllowZeroBatches allows the file to have zero batches
 	AllowZeroBatches bool `json:"allowZeroBatches"`
+
+	// BypassCompanyIdentificationMatch allows batches in which the Company Identification field
+	// in the batch header and control do not match.
+	BypassCompanyIdentificationMatch bool `json:"bypassCompanyIdentificationMatch"`
 }
 
 // ValidateWith performs NACHA format rule checks on each record according to their specification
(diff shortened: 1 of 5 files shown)


file.go
```diff
@@ -564,6 +564,9 @@ type ValidateOpts struct {
 
 	// AllowZeroBatches allows the file to have zero batches
 	AllowZeroBatches bool `json:"allowZeroBatches"`
+
+	// BypassCompanyIdentificationMatch can be set to skip validation for //TODO
+	BypassCompanyIdentificationMatch bool `json:"bypassCompanyIdentificationMatch"`
```
vxio, line 570: Originally I tried 
```go
	// ...
	BatchControl struct {
		BypassCompanyIdentificationMatch bool
	}
```
But an embedded struct makes the API awkward to use since you have to set it using 
```go
	batcher.SetValidation(&ValidateOpts{
		BatchControl: struct{
			BypassCompanyIdentificationMatch bool,
		}{
		BypassCompanyIdentificationMatch: true,
		}
	})
```

adamdecaf: That's a good point. I'm okay calling it *Match.

adamdecaf (approved):

jawspeak (commented):
Thanks very much.

vxio:
> Thanks very much.

Sure thing, thanks again for linking the use-case!

PR #907, opened 2021-04-27
Update documentation for Read, Create, Validate for clarity

### LINKED ISSUE #902: file : update documentation for Read(), Create(), and Validate()
ACH Version: `v1.7.1`

In the Go library, for `File` methods like `Read(), Create(), and Validate()`, the behavior and the types of errors that can be returned by each method is not clear to users (and whether errors overlap between these methods).

We should update the code comments (GoDoc) to make this more explicit.

Public slack convo for reference: https://moov-io.slack.com/archives/CD9J8EJKX/p1618875926139000

vxio:
Fixes #902

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -37,7 +37,7 @@ var (
 	maxLines = 2 + 2000000 + 100000000 + 8
 )
 
-// Reader reads records from a ACH-encoded file.
+// Reader reads records from an ACH-encoded file.
 type Reader struct {
 	// file is ach.file model being built as r is parsed.
 	File File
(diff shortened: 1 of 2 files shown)


reader.go
```diff
@@ -120,11 +120,14 @@ func NewReader(r io.Reader) *Reader {
...
 
-// Read reads each line of the ACH file and defines which parser to use based on the first character
-// of each line. It also enforces ACH formatting rules and returns the appropriate error if issues are found.
+// Read reads each line in the underlying io.Reader and returns a File and any errors encountered.
 //
-// A parsed file may not be valid and callers should ensure the file is validate with Validate()
-// and tabulate the file with Create(). Invalid files may be rejected by other Financial Institutions or ACH tools.
+// Read enforces ACH formatting rules and the first character of each line determines which parser is used.
+//
+// The returned File may not be valid. Callers should tabulate the File with File.Create followed by
```
vxio, line 127: Following the comment styles from the Go Standard library, I'm not using `()` when referencing functions. Is it fine if we start conforming to this convention?

nlakritz: I think we should, looks like we already do in a lot of comments. @adamdecaf, what do you think?  I can go through and remove all the `()`s if we decide to go that route.

nlakritz (approved):

reader.go
adamdecaf: Removing `()`'s is fine with me.

adamdecaf (approved):

PR #908, opened 2021-04-27
Limit file entry hash to 10 digits

### LINKED ISSUE #905: file: force a 10-digit entry hash in File.Create()
ACH Version: `v1.7.1`

**Context**
Since a call to `File.Create()` should assemble the file to a valid state, we want to prevent a scenario where the file's entry hash overflows to 11-digits.

**Solution**
In `File.Create()`, we can force this behavior by truncating the entry hash to 10-digits when it is set.

vxio:
Fixes #905

### DIFF
--- a/cmd/readACH/main.go
+++ b/cmd/readACH/main.go
@@ -56,20 +56,19 @@ func main() {
 		log.Panicf("Can not open file: %s: \n", err)
 	}
 
-	r := ach.NewReader(f)
-	achFile, err := r.Read()
+	achFile, err := ach.NewReader(f).Read()
 	if err != nil {
 		fmt.Printf("Issue reading file: %+v \n", err)
-	}
-
-	// ensure we have a validated file structure
-	if achFile.Validate(); err != nil {
-		fmt.Printf("Could not validate entire read file: %v", err)
 	}
 
 	// If you trust the file but it's formating is off building will probably resolve the malformed file.
 	if err := achFile.Create(); err != nil {
 		fmt.Printf("Could not create file with read properties: %v", err)
+	}
+
+	// ensure we have a validated file structure
+	if err := achFile.Validate(); err != nil {
+		fmt.Printf("Could not validate entire read file: %v", err)
 	}
 
 	// Output file contents
--- a/file.go
+++ b/file.go
@@ -22,7 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"math"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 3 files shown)


file.go
```diff
@@ -750,7 +754,9 @@ func (f *File) calculateEntryHash(IsADV bool) int {
 			hash = hash + batch.GetADVControl().EntryHash
 		}
 	}
-	return hash
+
+	// Ensure the entry hash cannot exceed 10 digits
+	return hash % int(math.Pow(10, 10))
```
adamdecaf: We do this inside each `Batch` with `numericField` so we could do that again instead of this conversion. Thoughts?

https://github.com/moov-io/ach/blob/master/batch.go#L770-L776

adamdecaf: `numericField` does give us the result we're looking for. 
https://github.com/moov-io/ach/blob/master/converters_test.go#L91

adamdecaf:
👍  Can we add a test for this?

file.go
vxio: I was kind of curious about performance. I'm fine with using the latter approach used in `Batch`.
```go
//pkg: github.com/moov-io/ach
//cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
//BenchmarkModulo
//BenchmarkModulo-16    	1000000000	         0.2669 ns/op
//PASS
func BenchmarkModulo(b *testing.B) {
	x := int(math.Pow10(10))
	for i := 0; i < b.N; i++ {
		_ = i % x
	}
}

//pkg: github.com/moov-io/ach
//cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
//BenchmarkStringCov
//BenchmarkStringCov-16    	10427030	       117.8 ns/op
//PASS
func BenchmarkStringCov(b *testing.B) {
	converters := &converters{}
	for i := 0; i < b.N; i++ {
		s := converters.numericField(i, 10)
		_, _ = strconv.Atoi(s)
	}
}

```

vxio:
> 👍 Can we add a test for this?

Sure thing

file.go
vxio: If anything, in a separate PR, I can create a helper function that'll take in an `(v int, maxDigits uint)` and return an `int` that can be used across Files and Batches; and from there we can implement the optimal conversion.

adamdecaf: Totally agree. We should truncate `int` values and then later convert them to strings.

adamdecaf (dismissed):

cmd/readACH/main.go
```diff
@@ -56,22 +56,21 @@ func main() {
...
 
-	r := ach.NewReader(f)
-	achFile, err := r.Read()
+	achFile, err := ach.NewReader(f).Read()
 	if err != nil {
 		fmt.Printf("Issue reading file: %+v \n", err)
 	}
 
-	// ensure we have a validated file structure
-	if achFile.Validate(); err != nil {
```
adamdecaf, line 66: FYI I see this mistake in a couple other files -- https://github.com/search?q=org%3Amoov-io+%22if+achFile%22&type=code

adamdecaf (approved):

PR #912, opened 2021-04-29
writer: add an option to bypass file validation

### LINKED ISSUE #903: writer: add an option to bypass call to File.Validate()
ACH Version: `1.7.1`

Write an invalid ACH file to an underlying writer using `Write(file *File)` 

The operation to succeed with no error.

**What did you see?**
The call to `Write(..)` will fail because of a call to `file.Validate()`.

**One use-case would be a system that:**
- reads in an invalid ACH file (`Reader.Read(..)` does not make a call to `File.Validate()`)
- makes minor adjustments like updating the immediate origin
- writes the invalid file out to be forwarded to another system

vxio:
Fixes #903

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -23,15 +23,14 @@ import (
 	"strings"
 )
 
-// A Writer writes an ach.file to a NACHA encoded file.
-//
-// As returned by NewWriter, a Writer writes ach.file structs into
-// NACHA formatted files.
-//
+// Writer writes a File to an io.Writer.
+// The File is validated against Nacha guidelines unless BypassValidation is enabled.
 type Writer struct {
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -23,15 +23,14 @@ import (
 	"strings"
 )
 
-// A Writer writes an ach.file to a NACHA encoded file.
-//
-// As returned by NewWriter, a Writer writes ach.file structs into
-// NACHA formatted files.
-//
+// Writer writes a File to an io.Writer.
```
vxio, line 26: Updated the GoDoc here for clarity. The previous code comments were a bit redundant.

adamdecaf (approved):

PR #914, opened 2021-04-30
add test example that causes entry details to be grouped into a single batch

vxio:


adamdecaf:
I think https://github.com/moov-io/ach/pull/916 has replaced this, right?

vxio:
> I think #916 has replaced this, right?

Yep, closing this, thank you!

PR #916, opened 2021-04-30
reader: chomp currentBatch whenever a BatchControl is encountered

### LINKED ISSUE #915: reader: invalid BatchControl records cause incorrect Batch accumulation
ACH Version: `v1.7.1`

Reading a file with invalid BatchControl records will cause the Reader to improperly accumulate Batches/Entry records. The Reader will not close off the `currentBatch` and instead subsequent EntryDetail records are included in the incorrect Batch.

A parsed `File` should represent the Nacha file as accurately as possible. 

**What did you see?**
An incorrect `File` with missing Batches and EntryDetail records in the wrong Batch. 

**How can we reproduce the problem?**

adamdecaf:
Previously a BatchControl record needed to be valid in order for the currentBatch to be properly accumulated into the File. This can produce invalid Files by including EntryDetail records from a later batch in a previous batch.

Fixes: https://github.com/moov-io/ach/issues/915


test/issues/issue915_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+	file, _ := rdr.Read()
+
+	if n := len(file.Batches); n != 12 {
+		t.Errorf("found %d batches", n)
+	}
+	for i := range file.Batches {
+		entries := file.Batches[i].GetEntries()
+		if n := len(entries); n != 1 {
+			t.Errorf("Batch[%d] has %d batches", i, n)
```
vxio:
```suggestion
			t.Errorf("Batch[%d] has %d entries", i, n)
```

vxio (dismissed):

vxio (approved):

Issue #922, opened 2021-05-05
achcli parse errors are unhelpful for large files


5outh:
## What I'm trying to do

I am reading the JSON version of a large (2.6MB) ACH file into an application, using `achcli` as a bridge. This works fine for most files, but when something is wrong, `achcli` explodes without any guidance about what is wrong:

```shell
$ achcli -v -reformat json big-ach.ach
moov-io/ach:v1.5.2 cli tool
found 1 ACH files to describe: big-ach.ach
ERROR: unable to read big-ach.ach
```

## What I want to see

More information about which rows were invalid. In this case, it's nearly impossible to take action against this error; I'm just stuck!

## Other notes

Describing the file with `achcli big-ach.ach` works completely fine. Maybe validation is failing? Possibly related to https://github.com/moov-io/ach/issues/828

adamdecaf:
Hey @5outh thanks for the bug report! I see we aren't displaying the error in that case, so I opened https://github.com/moov-io/ach/pull/924 to return the error information.

Here's an example: 
```
$ achcli -reformat json test/testdata/invalid-two-micro-deposits.ach
ERROR: unable to read test/testdata/invalid-two-micro-deposits.ach:
 line:17 record:Batches *ach.BatchError batch #2 (PPD) TotalDebitEntryDollarAmount calculated 0 is out-of-balance with batch control 44
```

Does that look better to you? FYI I see you're on v1.5.2 and we've released a few versions from there that might help to parse invalid files.

5outh:
Nice, yes that helps a lot! There are a number of problems with this file, and it would be nice long-term to get them _all_ up front if that's possible.

Thanks for the speedy turnaround!

adamdecaf:
@vxio are you okay if I release v1.8.1 with this PR? https://github.com/moov-io/ach/pull/925 might want to be its own release.

vxio:
@adamdecaf sure thing!

adamdecaf:
@5outh v1.8.1 has been released. Can you try it out?

https://github.com/moov-io/ach/releases/tag/v1.8.1

PR #924, opened 2021-05-05
cmd/achcli: include last error message in -reformat

adamdecaf:
Issue: https://github.com/moov-io/ach/issues/922

### DIFF
--- a/cmd/achcli/reformat.go
+++ b/cmd/achcli/reformat.go
@@ -42,13 +42,15 @@ func reformat(as string, filepath string) error {
 }
 
 func readIncomingFile(path string) (*ach.File, error) {
-	if file, err := readJsonFile(path); file != nil && err == nil {
+	file, err := readJsonFile(path)
+	if file != nil && err == nil {
 		return file, nil
 	}
-	if file, err := readACHFile(path); file != nil && err == nil {
+	file, err = readACHFile(path)
+	if file != nil && err == nil {
 		return file, nil
 	}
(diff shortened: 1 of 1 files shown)


cmd/achcli/reformat.go
```diff
@@ -42,13 +42,15 @@ func reformat(as string, filepath string) error {
...
+	if file != nil && err == nil {
 		return file, nil
 	}
-	if file, err := readACHFile(path); file != nil && err == nil {
+	file, err = readACHFile(path)
+	if file != nil && err == nil {
 		return file, nil
 	}
-	return nil, fmt.Errorf("unable to read %s", path)
+	return nil, fmt.Errorf("unable to read %s:\n %v", path, err)
```
vxio, line 53:
```suggestion
	return nil, fmt.Errorf("unable to read %s:\nError: %v", path, err)
```
Not necessary, but could help readability

vxio (approved):

cmd/achcli/reformat.go
adamdecaf: This seems to be already done by the `err`:

```
$ achcli -reformat json test/testdata/invalid-two-micro-deposits.ach
ERROR: unable to read test/testdata/invalid-two-micro-deposits.ach:
 line:17 record:Batches *ach.BatchError batch #2 (PPD) TotalDebitEntryDollarAmount calculated 0 is out-of-balance with batch control 44
```

vxio: ah you're right, thank you

PR #926, opened 2021-05-07
Set batchID from header on POST to create file

### LINKED ISSUE #923: server: batch ID not set on POST to /files/create
ACH Version: `v1.8.0`

Run the following commands against the server:
```bash
curl -X POST -H "content-type: application/json" -d "@ppd-valid.json" localhost:8080/files/create
// no error

curl localhost:8080/files/adam-01/batches/adam-01
// not found error
```
Successfully fetching the batch by its ID.

**Cause**
Only `POST /files/{fileID}/batches` sets the batch ID from its header; but `POST /files/create` doesn't. So the batch ID remains empty.

vxio:
Fixes #923

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,6 +84,7 @@ func NewBatchMTE(bh *BatchHeader) *BatchMTE {
 	batch := new(BatchMTE)
 	batch.setControl(NewBatchControl)
 	batch.SetHeader(bh)
+	batch.SetID(bh.ID)
 	return batch
 }
 ```
--- a/batchACK.go
+++ b/batchACK.go
@@ -29,6 +29,7 @@ func NewBatchACK(bh *BatchHeader) *BatchACK {
 	batch := new(BatchACK)
 	batch.SetControl(NewBatchControl())
 	batch.SetHeader(bh)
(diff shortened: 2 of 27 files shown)


batchACK.go
```diff
@@ -29,6 +29,7 @@ func NewBatchACK(bh *BatchHeader) *BatchACK {
 	batch := new(BatchACK)
 	batch.SetControl(NewBatchControl())
 	batch.SetHeader(bh)
+	batch.SetID(bh.ID)
```
vxio, line 32: I don't love setting this here, but following the existing pattern

CONTRIBUTING.md
```diff
@@ -84,6 +84,7 @@ func NewBatchMTE(bh *BatchHeader) *BatchMTE {
 	batch := new(BatchMTE)
 	batch.setControl(NewBatchControl)
 	batch.SetHeader(bh)
+    batch.SetID(bh.ID)
```
adamdecaf: Formatting?

vxio: dang, I thought I fixed this, one sec

adamdecaf (dismissed):

adamdecaf (approved):

Issue #927, opened 2021-05-10
MergeFiles combines conflicting TraceNumbers 


adamdecaf:
ACH Version: `v1.8.1`

**What were you trying to do?**
Calling MergeFiles on a directory of files should produce valid Nacha files.

**What did you see?**
Conflicting TraceNumbers are combined in the same file which causes FlattenBatches to fail.

PR #931, opened 2021-05-13
merge: combine entries together when Batch Headers match

adamdecaf:
Issue: https://github.com/moov-io/ach/issues/927

### DIFF
--- a/batchHeader.go
+++ b/batchHeader.go
@@ -25,8 +25,6 @@ import (
 	"unicode/utf8"
 )
 
-// msgServiceClass
-
 // BatchHeader identifies the originating entity and the type of transactions
 // contained in the batch (i.e., the standard entry class, PPD for consumer, CCD
 // or CTX for corporate). This record also contains the effective date, or desired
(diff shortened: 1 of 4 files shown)


batchHeader.go
```diff
@@ -213,6 +211,30 @@ func (bh *BatchHeader) String() string {
...
+	}
+	if bh.StandardEntryClassCode != other.StandardEntryClassCode {
+		return false
+	}
+	if bh.EffectiveEntryDate != other.EffectiveEntryDate {
+		return false
+	}
+	if bh.ODFIIdentification != other.ODFIIdentification {
+		return false
+	}
```
vxio, line 234:
```suggestion
	if bh.ServiceClassCode != other.ServiceClassCode &&
	!strings.EqualFold(bh.CompanyName, other.CompanyName) &&
	bh.CompanyIdentification != other.CompanyIdentification &&
	bh.StandardEntryClassCode != other.StandardEntryClassCode &&
        bh.EffectiveEntryDate != other.EffectiveEntryDate &&
	bh.ODFIIdentification != other.ODFIIdentification {
		return false
	}
```

merge.go
```diff
@@ -151,13 +163,27 @@ func (fs *mergableFiles) findOutfile(f *File) *File {
...
 }
 
+func mergeEntries(b1, b2 Batcher) Batcher {
+	b, _ := NewBatch(b1.GetHeader())
+	entries := sortEntriesByTraceNumber(append(b1.GetEntries(), b2.GetEntries()...))
+	for i := range entries {
+		b.AddEntry(entries[i])
+	}
+	b.SetControl(b1.GetControl())
+	b.Create()
```
vxio: This should never err right?

vxio (dismissed):
Had minor comments, but lgtm

batchHeader.go
adamdecaf: I was debating this. Putting it all into one `if` seems a bit messy. We don't really do that elsewhere in this project.

vxio: I'm fine with leaving as is

merge.go
adamdecaf: Correct. We can check that error - probably should.

vxio (approved):

PR #946, opened 2021-06-23
Update browser demo description in readme

nlakritz:
We now support bidirectional conversion in https://oss.moov.io/ach/


vxio (approved):
Noice!

PR #960, opened 2021-07-19
Keep settlement date when parsing files.

eweise:


batchHeader.go
```diff
@@ -179,8 +179,9 @@ func (bh *BatchHeader) Parse(record string) {
 	// 70-75 Date transactions are to be posted to the receivers' account.
 	// You almost always want the transaction to post as soon as possible, so put tomorrow's date in YYMMDD format
 	bh.EffectiveEntryDate = bh.validateSimpleDate(record[69:75])
-	// 76-79 Always blank (just fill with spaces)
-	bh.settlementDate = "   "
+	// 76-78 Always blank if creating batches (just fill with spaces).
+	// Set to file value when parsing. Julian day format.
+	bh.settlementDate = bh.validateSettlementDate(record[75:78])
```
adamdecaf, line 184: Did you want to export this value? Or is your use-case just to read and write the Settlement Date?

eweise: Good question. I think we would want to get the value at least. Updated the settlementDateField to be public

adamdecaf (dismissed):

validators.go
```diff
@@ -496,3 +496,28 @@ func CheckRoutingNumber(routingNumber string) error {
...
+	if s == emptyField {
+		return s
+	}
+
+	if len(s) != len(emptyField) {
+		return emptyField
+	}
+
+	day, err := strconv.Atoi(s)
+
```
vxio:
```suggestion
```

validators.go
```diff
@@ -496,3 +496,28 @@ func CheckRoutingNumber(routingNumber string) error {
...
+func (v *validator) validateSettlementDate(s string) string {
+	emptyField := "   "
+
+	if s == emptyField {
+		return s
+	}
+
+	if len(s) != len(emptyField) {
+		return emptyField
+	}
```
vxio:
```suggestion
	if s == emptyField || len(s) != len(emptyField) {
		return emptyField
	}
```

vxio (dismissed):
Two minor suggestions, but looks good!

adamdecaf (approved):

vxio (approved):

vxio:
Appreciate the PR @eweise , I'm going to merge this in!

eweise:
Awesome. Thank you both

PR #966, opened 2021-07-22
Implementing CustomReturnCode validation

### LINKED ISSUE #964: validation: support custom return codes on return entries
ACH Version: `1.93.`

Attempted to parse a return with a `R97` code. `R97` doesn't seem to exist, according to my 2021 Nacha operating rules & guidelines, so I'm thinking it might be an older code. 

I would love if return addenda (99 specifically) had custom validation on it that allowed custom return codes

emarcey:
Implementing new custom validator `customReturnCode` that allows returns with non-standard return codes to be parsed successfully.

Fixes https://github.com/moov-io/ach/issues/964


file.go
```diff
@@ -594,6 +594,10 @@ type ValidateOpts struct {
 	// BypassCompanyIdentificationMatch allows batches in which the Company Identification field
 	// in the batch header and control do not match.
 	BypassCompanyIdentificationMatch bool `json:"bypassCompanyIdentificationMatch"`
+
+	// CustomReturnCodes can be set to skip validation for the Return Code field in an Addenda99
+	// This allows for non-standard/deprecated return codes (e.g. R97)
+	CustomReturnCodes bool `json:"customReturnCodes"`
```
adamdecaf, line 600: @vxio What are your thoughts on a `CheckReturnCode func(code string) error` function that allows validating codes like "R97" as the issue mentions.

emarcey: For ref: https://moov-io.slack.com/archives/CD9J8EJKX/p1626989492134600?thread_ts=1626981488.133100&cid=CD9J8EJKX

I went with the bool so that I can call from an API endpoint more easily with the base docker image

adamdecaf: Yea that's a good idea. We can add a validation func in a separate PR.

adamdecaf (approved):

PR #975, opened 2021-08-09
batch: allow unequal ServiceClassCodes to be used

### LINKED ISSUE #973: Add ability to skip validations for NOC records
ACH Version: `v1.10.0`

When we call `/file/create` endpoint with an ach file consists of NOC records, we encounter validation errors specifically:
`ServiceClassCode header 225 is not equal to control 200`, `TransactionCode service class code 225 does not support transaction code 21`

It would be great if we could have the option to skip NOC validations.

We expected the endpoint to succeed and calling `/file/{file_id}` would return the parsed ach file.

**What did you see?**
Validation errors, although we are processing NOC records only.

**How can we reproduce the problem?**

Using an ach file containing the following info
```
5225 Company (Batch Header Service class code)
621( Entry Detail Record Transaction code)
8200 (Batch Control Service Class Code```

### PR DESCRIPTION
**Pull request by adamdecaf**
Fixes: https://github.com/moov-io/ach/issues/973

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -316,7 +316,8 @@ func (batch *Batch) verify() error {
 
 	if !batch.IsADV() {
 		// validate batch header and control codes are the same
-		if batch.Header.ServiceClassCode != batch.Control.ServiceClassCode {
+		if (batch.validateOpts == nil || !batch.validateOpts.UnequalServiceClassCode) &&
+			batch.Header.ServiceClassCode != batch.Control.ServiceClassCode {
 			return batch.Error("ServiceClassCode",
 				NewErrBatchHeaderControlEquality(batch.Header.ServiceClassCode, batch.Control.ServiceClassCode))
 		}
@@ -338,7 +339,8 @@ func (batch *Batch) verify() error {
 				NewErrBatchHeaderControlEquality(batch.Header.BatchNumber, batch.Control.BatchNumber))
 		}
 	} else {
-		if batch.Header.ServiceClassCode != batch.ADVControl.ServiceClassCode {
+		if (batch.validateOpts == nil || !batch.validateOpts.UnequalServiceClassCode) &&
+			batch.Header.ServiceClassCode != batch.ADVControl.ServiceClassCode {
 			return batch.Error("ServiceClassCode",
 				NewErrBatchHeaderControlEquality(batch.Header.ServiceClassCode, batch.ADVControl.ServiceClassCode))
 		}
--- a/file.go
+++ b/file.go
@@ -598,6 +598,10 @@ type ValidateOpts struct {
 	// CustomReturnCodes can be set to skip validation for the Return Code field in an Addenda99
 	// This allows for non-standard/deprecated return codes (e.g. R97)
 	CustomReturnCodes bool `json:"customReturnCodes"`
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by adamdecaf** [comment]
@yongli-current could you look this over and try it out? I need to write a test using this new option still.

File: batch.go [hunk 1]
```diff
@@ -316,9 +316,11 @@ func (batch *Batch) verify() error {
 
 	if !batch.IsADV() {
 		// validate batch header and control codes are the same
-		if batch.Header.ServiceClassCode != batch.Control.ServiceClassCode {
-			return batch.Error("ServiceClassCode",
-				NewErrBatchHeaderControlEquality(batch.Header.ServiceClassCode, batch.Control.ServiceClassCode))
+		if batch.validateOpts == nil || !batch.validateOpts.UnequalServiceClassCode {
+			if batch.Header.ServiceClassCode != batch.Control.ServiceClassCode {
```
vxio: I think we can un-nest the if-statements with `&&` logic
```go
 if batch.Header.ServiceClassCode != batch.Control.ServiceClassCode &&
    (batch.ValidateOpts == nil || !batch.validateOpts.UnequalServiceClassCode) {
```

file.go
```diff
@@ -598,6 +598,10 @@ type ValidateOpts struct {
 	// CustomReturnCodes can be set to skip validation for the Return Code field in an Addenda99
 	// This allows for non-standard/deprecated return codes (e.g. R97)
 	CustomReturnCodes bool `json:"customReturnCodes"`
+
+	// UnequalServiceClassCode skips ensuring the ServiceClassCode in each pair of BatchHeader
+	// and BatchControl records match.
```
vxio:
```suggestion
	// UnequalServiceClassCode skips equality checks for the ServiceClassCode in each pair of BatchHeader
	// and BatchControl records.
```

batch.go
adamdecaf: Yea. It seemed easier read as multiple if statements though.

batch.go
```diff
@@ -338,9 +339,11 @@ func (batch *Batch) verify() error {
 				NewErrBatchHeaderControlEquality(batch.Header.BatchNumber, batch.Control.BatchNumber))
 		}
 	} else {
-		if batch.Header.ServiceClassCode != batch.ADVControl.ServiceClassCode {
-			return batch.Error("ServiceClassCode",
-				NewErrBatchHeaderControlEquality(batch.Header.ServiceClassCode, batch.ADVControl.ServiceClassCode))
+		if batch.validateOpts == nil || !batch.validateOpts.UnequalServiceClassCode {
+			if batch.Header.ServiceClassCode != batch.ADVControl.ServiceClassCode {
```
vxio: Could you update this block to match the above? My bad, I should've commented to track.

vxio (dismissed):

yongli-current:
does the new flag need to be added in here?
https://github.com/moov-io/ach/blob/c55bb58cbc39a260984ad96ebf82dfdfbae53857/server/files.go#L116:L134

adamdecaf:
Yes please!

yongli-current:
lgtm

vxio:
@adamdecaf could you add the new option to this file? https://github.com/moov-io/ach/blob/master/docs/create-file.md

vxio (approved):

PR #980, opened 2021-09-02
file: allow for missing file header and control

adamdecaf:
Also 
- file: unmarshal with ValidateOpts if set
- file: during creation check ValidateOpts for allowing empty headers


darwinz (approved):
Looks good to me

vxio (approved):
Looks good

PR #1002, opened 2022-03-28
feat: add support for "Dishonored Returns" with Addenda04 records

adamdecaf:


batch.go
```diff
@@ -973,7 +982,10 @@ func (batch *Batch) addendaFieldInclusionReturn(entry *EntryDetail) error {
 	if entry.Addenda98 != nil {
 		return batch.Error("Addenda98", ErrBatchAddendaCategory, entry.Category)
 	}
-	if entry.Addenda99 == nil {
+	if entry.Addenda04 == nil && entry.Addenda99 == nil {
+		if entry.Addenda04 == nil {
+			return batch.Error("Addenda04", ErrFieldInclusion)
+		}
```
alovak: am I right that line 989 (below) will never be run because `entry.Addenda04 == nil` is always true if we are inside?

adamdecaf: Yea.. That's a good point.

alovak (dismissed):

jawspeak:
We discussed in slack and it was my error for getting this term "addenda04". This is an addenda with addenda type code of `99`, just for the sub-record set of fields in a dishonored return. This means Adam may be renaming it something else such as `addenda99_dishonored_return.go` or another better name. Addenda04 doesn't exist (or doesn't apply)

addenda99_dishonored_return.go
```diff
@@ -0,0 +1,185 @@
...
+	Addenda99Dishonored := &Addenda99Dishonored{
+		recordType: "7",
+		TypeCode:   "99",
+	}
+	return Addenda99Dishonored
+}
+
+func (Addenda99Dishonored *Addenda99Dishonored) Parse(record string) {
+	if utf8.RuneCountInString(record) != 94 {
+		return
```
jawspeak, line 76: Thinking about the future, not this PR but it could be nice to have a logger available and log things like this as a warning (I think it would be a logged warning / but not something to error out and stop parsing).

Here it silently returns without doing anything. As teams process large volumes of files they periodically may see invalid lines in them, and surfacing visibility / warnings in logs could be useful. That said, this is a bigger change that probably would go across other places in the parsing elsewhere too.

addenda99_dishonored_return.go
```diff
@@ -0,0 +1,185 @@
...
+
+func (Addenda99Dishonored *Addenda99Dishonored) ReturnReasonCodeField() string {
+	return Addenda99Dishonored.stringField(Addenda99Dishonored.ReturnReasonCode, 2)
+}
+
+func (Addenda99Dishonored *Addenda99Dishonored) AddendaInformationField() string {
+	return Addenda99Dishonored.alphaField(Addenda99Dishonored.AddendaInformation, 21)
+}
+
+// TraceNumberField returns a zero padded TraceNumber string
```
jawspeak, line 190: I think this is really a space padded field, not zero padded (as the comment says)

addenda99_dishonored_return.go
adamdecaf: Yea I agree. There's room to improve this parsing and remove the silent errors.

addenda99_dishonored_return.go
adamdecaf: To keep the behavior matching the other Addenda99 `*Field()` calls here it'll be zero-value. The field directly will have the exact value from the source file.

PR #1041, opened 2022-07-07
fix: reduce memory usage during MergeFiles

adamdecaf:
We've been seeing getTraceNumbers use a ton of memory, which isn't surprising since it allocates (and then expands) an array of trace numbers. This can be bypassed with a mess of for loops instead.

Running achgateway's tests before:
Screen Shot 2022-07-07 at 10 57 02 AM
Screen Shot 2022-07-07 at 10 57 12 AM

After: 
Screen Shot 2022-07-07 at 10 57 26 AM
Screen Shot 2022-07-07 at 11 02 16 AM

### DIFF
--- a/entryDetail.go
+++ b/entryDetail.go
@@ -634,25 +634,3 @@ func sortEntriesByTraceNumber(entries []*EntryDetail) []*EntryDetail {
 	})
 	return entries
 }
-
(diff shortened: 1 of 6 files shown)


vxio (approved):
nice optimization

PR #1095, opened 2022-10-04
fix: copy forward the validate options when segmenting a file

yordis:


file.go
```diff
@@ -929,6 +929,11 @@ func (f *File) SegmentFile(_ *SegmentFileConfiguration) (*File, *File, error) {
 	creditFile := NewFile()
 	debitFile := NewFile()
 
+	if f.validateOpts != nil {
+		creditFile.validateOpts = f.validateOpts
```
adamdecaf: Can we call SetValidation instead?

adamdecaf (approved):

PR #1109, opened 2022-10-21
feat: add Reversal() to files

### LINKED ISSUE #756: file: offer a Reversal method
**Problem Describe**
If a duplicate or erroneous file was sent off to an FI it can be corrected by sending a reversal ACH file. This reversal has a specific transform from the original. These definitions come from the NACHA guidelines, but the process to create a reversal file is below.

Formatting Requirements for Reversing Entries:

> Must submit a Reversing Entry within a batch that includes the word “REVERSAL” in the Company Entry Description field of the Company/Batch Header Record.  The description “REVERSAL” must replace the original content of the Company Entry Description field transmitted in the original batch, including content that may otherwise have been required by The Rules.
> For each Reversing Entry, the content on the following fields must remain unchanged from the original, erroneous Entry to which the Reversal relates: 
>  - Standard Entry Class Code
>  - Company Identification/Originator Identification
>  - Amount

> Update the following records according to the fund flow
>   - Update TransactionCode fields. 
>   - Set EffectiveEntryDate properly (for same-day vs standard) 

> The name of the Originator (as identified in the Company Name field or the Originator Name field) must reflect the same Originator identified in the Erroneous Entry to which the Reversal relates.

adamdecaf:
Closes: https://github.com/moov-io/ach/issues/756

### DIFF
--- a/reversal.go
+++ b/reversal.go
@@ -1,0 +1,56 @@
+// Licensed to The Moov Authors under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. The Moov Authors licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package ach
+
+import (
+	"time"
+)
+
+// Reversal will transform a File into a Nacha compliant reversal which can be transmitted to undo fund movement.
+func (f *File) Reversal(effectiveEntryDate time.Time) error {
+	for i := range f.Batches {
+		bh := f.Batches[i].GetHeader()
+
+		// Must submit a Reversing Entry within a batch that includes the word "REVERSAL" in the
+		// Company Entry Description field of the Company/Batch Header Record.
+		//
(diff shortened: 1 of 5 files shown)


reversal.go
```diff
@@ -0,0 +1,56 @@
...
+// under the License.
+
+package ach
+
+import (
+	"time"
+)
+
+// Reversal will transform a File into a Nacha compliant reversal which can be transmitted to undo fund movement.
+func (f *File) Reversal(effectiveEntryDate time.Time) error {
```
atonks2, line 25: Would it be worth checking here `if effectiveEntryDate.IsZero()` and returning an error?

adamdecaf: Perhaps. That'll fail validation as `000000` isn't a valid EffectiveEntryDate value.

reversal.go
```diff
@@ -0,0 +1,56 @@
...
+		// Adjust Effective Entry Date for same-day vs standard
+		bh.EffectiveEntryDate = effectiveEntryDate.Format("060102")
+
+		f.Batches[i].SetHeader(bh)
+
+		// In EntryDetail records we need to update the TransactionCode fields to undo fund movement.
+		entries := f.Batches[i].GetEntries()
+		for j := range entries {
+			// 22 (CheckingCredit) -> 27 (CheckingDebit)
+			entries[j].TransactionCode += 5
```
atonks2: Is this always going to end up as `27`?, or is `TransactionCode + 5` always the correct code for a reversal?

adamdecaf: There are a few ranges of these codes, but it's correct to `+= 5` them all. 

https://github.com/moov-io/ach/blob/v1.21.2/entryDetail.go#L112

atonks2: Well that's convenient. Thanks!

atonks2 (dismissed):

reversal.go
adamdecaf: There's a bug with 27 -> 32 (which is incorrect).

atonks2 (dismissed):

PR #1111, opened 2022-10-24
chore: add testing about File Header Immediate Origin field

yordis:
## Notes

Trying to figure out why the `0` is removed from the final output. Is this always a thing? Should I expect this everywhere where I see the `bTTTTAAAAC` format?

### DIFF
--- a/examples/example_ppdWrite_segmentFile_test.go
+++ b/examples/example_ppdWrite_segmentFile_test.go
@@ -19,9 +19,11 @@ package examples
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"log"
 	"os"
 	"path/filepath"
+	"testing"
 
 	"github.com/moov-io/ach"
 )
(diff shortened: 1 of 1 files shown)


examples/example_ppdWrite_segmentFile_test.go
```diff
@@ -82,3 +84,13 @@ func Example_ppdWriteSegmentFile() {
...
 }
+
+func Test_ppdMixedDebitCreditImmediateOrigin(t *testing.T) {
+	f, err := os.Open(filepath.Join("testdata", "ppd-mixedDebitCredit.ach"))
+	require.NoError(t, err)
+	r := ach.NewReader(f)
+	achFile, err := r.Read()
+	require.NoError(t, err)
+
+	require.Equal(t, achFile.Header.ImmediateOrigin, "0121042882")
```
adamdecaf: You probably want to use the Field() accessors. 

```suggestion
	require.Equal(t, achFile.Header.ImmediateOriginField(), "0121042882")
```

adamdecaf:
It looks like we're removing the `0` to comply with post-2013 Nacha rules/guidelines. https://github.com/moov-io/ach/pull/652

PR #1121, opened 2022-10-28
fix: allow missing file headers and segmenting file

yordis:
## Context

Because the package combines `Parsing` + `Validating` it was not using the right `ach.ValidateOpts`

I had to first parse the options and then set the options to the empty file because when unmarshalling from JSON, you need the options to be there already https://github.com/moov-io/ach/blob/7837e7e9f2f1d1935bc1493b1e76249f55c710c2/file.go#L223

This unmarshal the options first, then assign to the file, and then do the parsing


adamdecaf:
Going with https://github.com/moov-io/ach/pull/1123 instead

PR #1128, opened 2022-11-08
fix: allow mixed-case Addenda10 TransactionType codes

### LINKED ISSUE #1126: Addenda Transaction Type is case sensitive but that is not a NACHA requirement
### ACH Version

1.18.5

Receiving an Addenda 10 record the transaction type can be lowercase. However, validation disallows lowercase entries.

https://github.com/moov-io/ach/blob/4b6da1cc19af79a153e1b9a98ce6b0159ea52c86/validators.go#L396L403

Lowercase Transaction Type codes to be allowed

### What did you see?

line:4 record:Addenda *ach.FieldError TransactionTypeCode mis is an invalid Addenda10 Transaction Type Code
  line:14 record:Batches *ach.BatchError batch #20 (IAT) FieldError Addenda10 <nil> is a mandatory field and has a default value

Process an Addenda 10 with a lower case `MIS` as the transaction type

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/1126

### DIFF
--- a/validators.go
+++ b/validators.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 	"unicode/utf8"
 )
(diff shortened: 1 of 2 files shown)


kmyers94 (commented):
LGTM! TY

PR #1149, opened 2022-12-17
fix: don't re-check whether the file is ADV while writing each batch

Deiz:
Re-computing file.IsADV() as each batch is written becomes remarkably expensive for files containing large number of batches.

This change just serves to compute it once up-front during each write.


adamdecaf (approved):

adamdecaf:
Yea, that's a great find. Thanks @Deiz!

PR #1158, opened 2023-01-08
feat: use testify's `assert` package for `Valid` unit tests

cpustejovsky:
* use `assert.True` and `assert.False` instead of `if !Valid` and `if Valid` as that seemed to be clearer.
* change packages to `iso4217_test` and `iso3166_test` to make sure only exported functions can be tested.

Let me know if these changes make sense. I saw the repo was already using testify's `require` package.

### DIFF
--- a/internal/iso3166/validate_test.go
+++ b/internal/iso3166/validate_test.go
@@ -15,26 +15,17 @@
 // specific language governing permissions and limitations
 // under the License.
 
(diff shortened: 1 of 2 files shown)


adamdecaf (approved):

adamdecaf:
Looks good, thanks @cpustejovsky!

Issue #1159, opened 2023-01-08
Feature Request: Preserve spacing when parsing ACH file


AndrewPecha:
### ACH Version

1.27.0

### What were you trying to do?

I was attempting to parse an entire ACH file to JSON using the /create POST endpoint and preserve the leading and trailing spaces for  `CompanyIdentification` in `batchHeader.go`.

Instead, a value of " 123456789" or "123456789 " evaluates to "123456789" with the leading or trailing space removed

### What did you expect to see?

A way to enable a parsing option to preserve spacing of certain elements that normally trim leading/trailing spaces when they are parsed

I know my use case of only caring about `CompanyIdentification` is pretty small, so would it make sense to create an option for the parsers to preserve these leading/trailing spaces where `strings.TrimSpace()` is normally called?

adamdecaf:
Yea, we could add a `PreserveSpaces bool` option. Could you open a pull request? It's not quite a ValidateOpt, but I don't want another struct (e.g. `ReaderOpts`).

AndrewPecha:
Fair enough, that makes sense to me too. I can get a PR submitted in the next couple days.

Thank you!

adamdecaf:
Fixed in https://github.com/moov-io/ach/pull/1162

PR #1162, opened 2023-01-12
Add 'PreserveSpaces' opt

AndrewPecha:
This is driven by [this issue from a few days ago](https://github.com/moov-io/ach/issues/1159)

- Added `SetValidation` to `BatchControl` so opts could be set before parsing
- Added `parseStringFieldWithOpts` to `converters` struct
- Modified parsing for `FileHeader`, `BatchHeader`, and `BatchControl`  to respect the new opt where spaces were normally trimmed
- Modified the parsing logic for `EntryDetail` to not trim spaces for RDFI, CheckDigit, and Trace Number (I believe these all need numeric values and use every character to be considered valid anyway, but please correct me if I'm wrong)


batchControl_test.go
```diff
@@ -315,3 +315,24 @@ func BenchmarkBatchControlLength(b *testing.B) {
...
+func TestBatchControl__SetValidation(t *testing.T) {
+	bh := NewBatchControl()
+	bh.SetValidation(&ValidateOpts{})
+
+	// nil out
+	bh = nil
+	bh.SetValidation(&ValidateOpts{})
+}
+
+func TestBatchControl__IsValidWithPreserveSpacesOpt(t *testing.T) {
```
AndrewPecha, line 328: Wanted to ensure nothing relied on spaces being trimmed when validating

entryDetail.go
```diff
@@ -204,9 +204,9 @@ func (ed *EntryDetail) Parse(record string) {
 	// 2-3 is checking credit 22 debit 27 savings credit 32 debit 37
 	ed.TransactionCode = ed.parseNumField(record[1:3])
 	// 4-11 the RDFI's routing number without the last digit.
-	ed.RDFIIdentification = ed.parseStringField(record[3:11])
+	ed.RDFIIdentification = record[3:11]
```
AndrewPecha, line 207: I think we are safe to drop the space trimming here altogether, but please correct me if I'm wrong

adamdecaf: Yea. It's probably fine with some fields.

adamdecaf (approved):

batchControl_test.go
adamdecaf: Thanks

PR #1178, opened 2023-02-20
Enhance FlattenBatches operation

ckbaum:
A few changes to the behavior of the `FlattenBatches` function of `File`:
- **Don't combine two batches if they share any trace numbers.** The NACHA spec indicates that trace numbers must be unique within a batch but not within the file, so currently we (correctly) see a validation error if two entries with the same trace number have similar batch headers and we call `FlattenBatches`. Instead, leave batches with similar headers separate if combining them would result in a trace number collision.
- **Make an effort to preserve original batch ordering.** Currently the way batch headers are loaded into a map as part of flattening means the batch order of the resulting file is random. Instead, have flattened batches ordered based on the lowest batch number of their contained batches. This makes the resulting file more closely resemble the original and makes the output of the flattening process deterministic, which is easier to test.

Because trace number collision prevention made the `FlattenBatches` code a bit more complicated:
- **Add more comprehensive tests.** Leave all the existing flatten tests intact, but also confirm that final entry groupings look exactly the way we expect, and that every entry detail is represented in the new file with the same associated header information.
- **Move batch flattening code to a dedicated file.** This allows us to use an interface for IAT and "other" batches, so the actual flattening logic is represented once, without making a big mess. Also makes for a convenient place to put the now-obnoxious batch flattening tests. But if this feels against the style of the repo happy to move this code back to `file.go`.

Two small fixture changes:
- `test/testdata/flattenADVBatchesMultipleBatchHeaders.ach` give each entry a different `Amount` so that they are easier to distinguish between the original and flattened files.
- `test/testdata/flattenIATBatchesMultipleBatchHeaders.ach` there were only 3 different trace numbers represented across 12 different entries, which made for a boring flattening.


adamdecaf:
First off thank you @ckbaum for making this PR. It addresses a lot of the shortcomings in the existing flatten routine, which have been needed for quite some time. I agree that maintaining as much of the original ordering as possible is ideal. 

Thanks for updating the tests to cover the code better. I [named some errors](https://github.com/moov-io/ach/pull/1178/commits/7cade0c2671d843c6e65620d5a13500a1a4b5931) and unified some of the code.

file_flattener.go
```diff
@@ -48,7 +54,9 @@ func FlattenedFile(originalFile *File) (*File, error) {
 
 	// Merge each original batch into a new batch
 	newBatchesByHeader := map[string][]mergeable{}
-	for _, batch := range originalBatches {
+	for i := range originalBatches {
+		batch := originalBatches[i]
```
adamdecaf: I've had issues in the past with the loop variable syntax. There are some open issues with Go to fix them, but I tend to discourage their use.

adamdecaf:
I'm happy to merge and release this if you are @ckbaum.

ckbaum:
Looks good to me!

file_flattener.go
```diff
@@ -0,0 +1,275 @@
...
+)
+
+// FlattenFile returns a flattened version of a File, where batches with similar batch headers are consolidated.
+//
+// Two batches are eligible to be combined if:
+//   - their headers match, excluding the batch number (which isn't used in return matching and reflects
+//     the final composition of the file.)
+//   - they don't contain any entries with common trace numbers, since trace numbers must be unique
+//     within a batch.
+func FlattenFile(originalFile *File) (*File, error) {
```
ckbaum: nit: This name (and `FlattenBatches`) make it sound a little like we're mutating the `File` rather than returning a flattened version. I actually thought `FlattenBatches` wasn't working when I first used it. Maybe it's just me though, not that important.

adamdecaf: Would `Flatten(file) (*File, error)` make more sense? You're right it sounds like a mutation operation.

ckbaum: `Flatten` is a bit better I think. `Flattened` or `FlattenedFile` feel the most to me like we are producing a new object, but I feel less conviction about it the more I think about it; this is a hard naming problem 😬 cool with whatever.

adamdecaf (approved):

PR #1225, opened 2023-05-11
fix: BatchHeaders are different with mismatching CompanyEntryDescription values

adamdecaf:
This comes up sometimes where customers put invoice numbers in this field and they see incorrect values in OLB memos.


vxio (dismissed):
Is there a unit test we could add for this use-case?

vxio (approved):

PR #1262, opened 2023-07-17
fix: allow full range of ASCII and EBCDIC characters accepted by Nacha

adamdecaf:
The 2022 Nacha rules state that Valid Characters are:

> 0-9, A-Z, a-z, space, EBCDIC values greater than hexadecimal "3F",
> ASCII values greater than hexadecimal "1F".

- [ASCII Characters](https://en.wikipedia.org/wiki/ASCII#Character_set)
- [EBCDIC Characters](https://en.wikipedia.org/wiki/EBCDIC)

### DIFF
--- a/validators_test.go
+++ b/validators_test.go
@@ -175,6 +175,22 @@ func TestValidators__isAlphanumeric(t *testing.T) {
 	}
 }
 
+func TestValidators__isAlphanumericExamples(t *testing.T) {
+	v := validator{}
+
+	validCases := []string{`|`}
+	for i := range validCases {
+		err := v.isAlphanumeric(validCases[i])
+		require.NoError(t, err, fmt.Sprintf("input: %q", validCases[i]))
+	}
+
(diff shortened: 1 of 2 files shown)


validators_test.go
```diff
@@ -175,6 +175,22 @@ func TestValidators__isAlphanumeric(t *testing.T) {
 	}
 }
 
+func TestValidators__isAlphanumericExamples(t *testing.T) {
+	v := validator{}
+
+	validCases := []string{`|`}
```
alovak: I'm curious, is `|` alphanumeric (letters and numbers)? I guess it is in ACH. Right?

adamdecaf: `|` is `0x7C` [in ASCII](https://en.wikipedia.org/wiki/ASCII#Character_set) which is allowed by Nacha.

jasonbornsteinMOOV (approved):

wadearnold (approved):

PR #1304, opened 2023-10-16
feat: add a File iterator

adamdecaf:
This iterator will help us process files entry-by-entry which should have a large reduction of memory usage.

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -1,0 +1,139 @@
+// Licensed to The Moov Authors under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. The Moov Authors licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package ach
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+)
+
+// Iterator is a data structure for processing an ACH file one entry at a time.
+type Iterator struct {
+	reader     *Reader
+	scanner    *bufio.Scanner
+	cachedLine string
+}
+
+// NewIterator returns an Iterator
+func NewIterator(r io.Reader) *Iterator {
+	reader := NewReader(strings.NewReader("")) // the input is not used, we rely on .readLine()
+	reader.skipBatchAccumulation = true        // don't call .AddBatch(..)
+
+	out := &Iterator{
+		reader:  reader,
+		scanner: bufio.NewScanner(r),
+	}
+	return out
+}
+
+func (i *Iterator) SetValidation(opts *ValidateOpts) {
+	if i.reader != nil {
+		i.reader.SetValidation(opts)
+	}
+}
+
+// GetHeader will return the FileHeader once encountered by the iterator.
+// Call NextEntry() at least once to populate the header.
+func (i *Iterator) GetHeader() *FileHeader {
+	if i.reader != nil {
+		return &i.reader.File.Header
+	}
+	return nil
+}
+
+// GetControl will return the FileControl once encountered by the iterator.
+// Call NextEntry() at least once to populate the control.
+func (i *Iterator) GetControl() *FileControl {
+	if i.reader != nil {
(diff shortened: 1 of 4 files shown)


adamdecaf:
cc @wadearnold

adamdecaf:
Including #1305 and this fix together gives a noticeable improvement over reading files fully in memory. 

The Basic_ tests are calling `ach.ReadFile(..)` and the Iterator tests are calling `NextEntry()` until they run out of entries. 

```
goos: darwin
goarch: amd64
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
Benchmark
Benchmark/Basic_TRANSACTIONS100223143314.TXT
Benchmark/Basic_TRANSACTIONS100223143314.TXT-16         	     100	 154561285 ns/op	13297808 B/op	  559303 allocs/op
Benchmark/Basic_TRANSACTIONS101023162208.TXT
Benchmark/Basic_TRANSACTIONS101023162208.TXT-16         	     134	 132010532 ns/op	 9418912 B/op	  482160 allocs/op
Benchmark/Iterator_TRANSACTIONS100223143314.TXT
Benchmark/Iterator_TRANSACTIONS100223143314.TXT-16      	     145	 122932698 ns/op	14206015 B/op	  448108 allocs/op
Benchmark/Iterator_TRANSACTIONS101023162208.TXT
Benchmark/Iterator_TRANSACTIONS101023162208.TXT-16      	     250	  71073532 ns/op	 8219774 B/op	  259784 allocs/op
```

adamdecaf:
The iterator does find EntryDetail records that are outside of a batch, so that's a benefit over ReadFile. I do have a patch for the iterator skipping blank lines and need to test it when blank lines are in the middle. That'll be easy tomorrow. 

<details>
<summary>Patch</summary>

```
@@ -84,6 +84,10 @@ func (i *Iterator) NextEntry() (*BatchHeader, *EntryDetail, error) {
 		i.scanner.Scan()
 		line = i.scanner.Text()
 		i.reader.lineNum++
+		if line == "" {
+			// TODO(adam): Can the reader and iterator handle newlines in the middle of a file?
+			return nil, nil, nil
+		}
 	}
 
 	if err := i.reader.readLine(line); err != nil {
@@ -102,6 +106,7 @@ func (i *Iterator) NextEntry() (*BatchHeader, *EntryDetail, error) {
 			for {
 				if i.scanner.Scan() {
 					foundLine := i.scanner.Text()
+					i.reader.lineNum++
 					if foundLine == "" {
 						break
 					}
```
</details>

wadearnold (approved):

InfernoJJ (approved):

iterator.go
```diff
@@ -0,0 +1,162 @@
...
+	reader := NewReader(strings.NewReader("")) // the input is not used, we rely on .readLine()
+	reader.skipBatchAccumulation = true        // don't call .AddBatch(..)
+
+	out := &Iterator{
+		reader:  reader,
+		scanner: bufio.NewScanner(r),
+	}
+	return out
+}
+
```
burnettekm, line 46: Do we have a style requirement to include a comment for each exported function?

iterator.go
```diff
@@ -0,0 +1,162 @@
...
+func (i *Iterator) NextEntry() (*BatchHeader, *EntryDetail, error) {
+	// Clear the reader's File
+	defer func() {
+		i.reader.File = File{}
+	}()
+
+	// Read the file one line at a time
+	line := i.cachedLine
+	if line != "" {
+		i.cachedLine = "" // clear cache
```
burnettekm, line 83: Why wouldn't a pre-existing cachedLine be an error state?

iterator.go
```diff
@@ -0,0 +1,162 @@
...
+			i.reader.lineNum++
+			if allSpaces(line) {
+				continue
+			}
+			if line != "" {
+				break
+			}
+		}
+		// If we've exhausted all lines in the reader then quit
+		if line == "" || allSpaces(line) {
```
burnettekm, line 97: Would this occur if the file is empty?

iterator.go
adamdecaf: No, but in the OSS code I try to include docs on most exported functions.

iterator.go
adamdecaf: EntryDetail records can have addendas which we want to include on the EntryDetail. Since a `bufio.Scanner` has no method to peek at the next `.Text()` result we have to cache the line if it's not an addenda. On the next iteration we want to take a cached line (and skip calling `.Text()` for that iteration) when it exists.

iterator.go
adamdecaf: Maybe? I would think the iterator would quit before this part of the code.

iterator.go
burnettekm: Thank you!

iterator.go
burnettekm: Makes sense. Thank you!

PR #1305, opened 2023-10-16
fix: process entry rune-by-rune without casting between string -> rune -> string

adamdecaf:
Converting between types is costly and can waste a lot of memory. Go lets us iterate through strings rune-by-rune, which helps to reduce some conversions.


adamdecaf:
cc @wadearnold

wadearnold (approved):

PR #1316, opened 2023-10-27
Parse file id modifier from file header; don't assume 'A'

ckbaum:
The file ID modifier won't always be "A"; other values are sometimes used to distinguish between multiple files with otherwise identical headers.

### DIFF
--- a/fileHeader.go
+++ b/fileHeader.go
@@ -118,68 +118,37 @@ func NewFileHeader() FileHeader {
 //
 // Parse provides no guarantee about all fields being filled in. Callers should make a Validate call to confirm successful parsing and data validity.
 func (fh *FileHeader) Parse(record string) {
-	runeCount := utf8.RuneCountInString(record)
-	if runeCount != 94 {
+	if utf8.RuneCountInString(record) != 94 {
 		return
 	}
+	runes := []rune(record)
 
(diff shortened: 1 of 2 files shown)


ckbaum:
@adamdecaf I'm a bit confused about the test failures I'm seeing here.

`test/issues/testdata/issue702-1.ach`, `test/issues/testdata/issue702.ach`, and `test/testdata/FISERV-ZEROFILE-PIMRET825324_032720_110221.ach` fixtures are all missing a file ID modifier. The rulebook is pretty clear that this character is required and must be uppercase A-Z or numeric 0-9. Not sure how these tests were passing before the recent change which hardcodes this value to `A`.

adamdecaf:
My bad on hardcoding this as part of the optimization. We were reading it out from the file before. 

I don't think we ever validated this field since it was mostly "A", but we'll start.

adamdecaf:
What do you think if we revert the minor perf optimization for FileHeader's and validate FileIDModifier?

ckbaum:
Either way sounds fine to me! Thanks 👍

adamdecaf:
https://github.com/moov-io/ach/releases/tag/v1.33.2 has been released with this, thanks @ckbaum!

PR #1335, opened 2023-12-07
Update amount to be long to allow handling of amounts larger than 21M

choudharynitu:


adamdecaf (dismissed):

adamdecaf:
Thanks for the fix!

openapi.yaml
```diff
@@ -794,11 +794,11 @@ components:
           example: 0
         totalDebit:
           description: Accumulated Batch debit totals within the file.
-          type: integer
+          type: long
```
adamdecaf: I think we need to specify `format: int64` and keep `type: integer`. 

See: https://swagger.io/docs/specification/data-models/data-types/#numbers

choudharynitu: let me fix it, when I generated a java client from this yaml file it worked fine, did not realize need to add the format

adamdecaf (approved):

PR #1336, opened 2023-12-07
Better wrap text output in textarea for webui

nickyost:
Enable better wrapping for the output textarea to ensure no auto-wrapping occurs, which makes the "human readable" output much harder to read without copy/pasting the text into another editor


adamdecaf (approved):

adamdecaf:
Thanks for the fix!

PR #1357, opened 2024-01-20
Use constructor to create IATEntryDetails when parsing IAT batches

ckbaum:
- Use `NewIATEntryDetail()` (rather than `new(IATEntryDetail)`) to create IAT entry details when parsing a file; this sets a couple important defaults including `Category`, which is currently blank for all parsed IAT forwards.
- Strip extra whitespace by default when parsing account numbers of IAT entries, to match the behavior of parsed account numbers for other entry details (we can leave this change off if it feels potentially disruptive.)

### DIFF
--- a/cmd/achcli/describe/file_test.go
+++ b/cmd/achcli/describe/file_test.go
@@ -35,7 +35,7 @@ func TestDescribeIAT(t *testing.T) {
 	if testing.Verbose() {
 		os.Stdout.Write(buf.Bytes())
 	}
-	require.Equal(t, 5255, buf.Len())
+	require.Equal(t, 5262, buf.Len())
 }
 
 func TestDescribeReturn(t *testing.T) {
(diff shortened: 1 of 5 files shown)


reader_test.go
```diff
@@ -1883,18 +1895,24 @@ func TestCategoryAssignment(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	categoriesByDiscretionaryData := map[string]string{}
+	categoriesByAccountNumber := map[string]string{}
```
ckbaum, line 1898: switched this test to look up entries by account number, since IAT entries don't have discretionary data

cmd/achcli/describe/file_test.go
```diff
@@ -35,7 +35,7 @@ func TestDescribeIAT(t *testing.T) {
 	if testing.Verbose() {
 		os.Stdout.Write(buf.Bytes())
 	}
-	require.Equal(t, 5255, buf.Len())
+	require.Equal(t, 5262, buf.Len())
```
ckbaum, line 38: now includes "Forward"

adamdecaf (approved):
Changes look good. Thank you!

PR #1377, opened 2024-02-25
Correctly assign Category to IAT entries

ckbaum:
`Category` isn't currently being set correctly for IAT returns and NOCs. Add the assignment and a couple new tests.


adamdecaf (approved):
Thanks for the fix!

adamdecaf:
Released in https://github.com/moov-io/ach/releases/tag/v1.34.3

PR #1380, opened 2024-02-27
Generate jni bindings

midepeter:
This PR is to realated to initial !1327 for generating java bindings for the ach library


adamdecaf:
Can you include only the changes for JNI interop in this PR?

javainterop/nextEntry.go
```diff
@@ -0,0 +1,29 @@
+package javainterop
```
adamdecaf: Can we put all of this in one `javainterop.go` file that has an alias for every type? We can figure out a script or something to generate the file if that's easier.

midepeter:
Yes, that was what I intended to do ..

I think I had some commit issues … not sure why the commits mixed up but I fixed it up manually.

javainterop/nextEntry.go
midepeter: that javainterop.go should have an alias for every type in the ach package right

are we porting every function in the ach package to the javainterop package or just the ones we are trying to reformat (SegmentFile and NextEntry)

adamdecaf: Yes it should have an alias to every type and function so they're available to someone using Java. We need to offer wrappers for some of the functions that don't compile nicely.

midepeter: Oh alright.

midepeter: @adamdecaf 

I am having a `infinite recursive call` lint issue when calling the `ach` functions in the `javainterop` functions. I don't think this affects the gobind generation but should i be mindful of that?

adamdecaf: Can you share the full error?

midepeter: `infinite recursive call (SA5007)go-staticcheck`

It is not a full error. Just a warning for the go-staticcheck

adamdecaf: Are you trying to generate the JNI code for the entire moov-io/ach repository? You should be able to target the `javainterop` package as the only thing to compile/generate.

javainterop/nextEntry.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func (i *Iterator) NextEntry(r io.Reader) NextEntryResponse {
+	batchHeader, entryFile, err := ach.NewIterator(r).NextEntry()
+
+	return NextEntryResponse{
+		NextBatchHeader: BatchHeader(*batchHeader),
+		NextEntryDetail: EntryDetail(*entryFile),
+		NextEntryError:  err,
+	}
```
adamdecaf: We will want to make the `Iterator` a struct holding a reference to the underlying `ach.Iterator` so that each call to `NextEntry()` will progress record by record. What you have now wouldn't quite work.

midepeter: Yes, this is true.

javainterop/nextEntry.go
midepeter: No, I am targeting just the `javainterop` package.

Actually, I realized if we go the route of referencing the ach types by embedding them into structs and calling the exported functions on them like you highlighted for the `Iterator` and `NextEntry` below. It eliminates this error. I think it is because of the `alias` as it is, does not necessarily create a newType but just masks the `oldType`

adamdecaf: Ah that could be. Sounds good!

midepeter:
@adamdecaf 

could you help take a look at the update. Thank you

midepeter:
@adamdecaf 

Trust you are doing great.

Hopefully this passes the pending checks but i need a review from you on what i did to know if it fits for merging or if there is something more to do.

Thank you.

midepeter:
@adamdecaf 

The ubuntu build is failing due to the test coverage. I though since we were porting the library the unit testing was not compulsory. I am not sure what I should do in this case.

jason53:
@adamdecaf were these changes never merged in?  i am just following up from our discussion.  https://github.com/moov-io/ach/discussions/1373#discussioncomment-8619567

adamdecaf:
Correct they were not merged in. We still need to figure out a way to setup types/classes so the JNI layer supports them all. 

I've tried another way but there are tons of unsupported classes, which is annoying.  See https://github.com/moov-io/ach/pull/1406

PR #1388, opened 2024-03-01
fix: parse large files without newlines

### LINKED ISSUE #1381: Reader cannot consume large, single-line ACH files
### ACH Version

master

We sometimes receive large ACH files where: 

1. the individual 94 byte records are not separated by newlines
2. the entire file exceeds the `bufio.MaxTokenSize`

 `ach.Reader` runs into problems parsing the file because it uses a `Scanner` that splits on newlines, and has a 64k maximum buffer/token size.

Because `ach.Reader.Read` doesn't actually check the result of `Scanner.Scan()` (or `Scanner.Err()`), it attempts to parse an arbitrary 64k chunk of data as an ACH record (or series of records) and fails.

`ach.Reader` should be able to consume arbitrarily large ACH files, with or without newline-delimited records (at least until we run out of memory).

I think using a custom `SplitFunc` (that breaks 94 byte records and/or newlines -- to handle "broken" lines that aren't padded to 94 characters) for the `Scanner` instead of the default `ScanLines` would be more reliable.

### What did you see?

ach.Reader fails to properly parse the file with some cryptic message about 

Attempt to read an ACH file that: 

1. Is not newline-delimited
2. Exceeds 64k bytes (`bufio.MaxScanTokenSize`)

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/1381


reader.go
```diff
@@ -179,14 +183,60 @@ func (r *Reader) Read() (File, error) {
...
+			if currentLineRuneCount > 0 {
+				goto fullLine
+			}
+		default:
+			currentLineRuneCount += 1
+			currentLine.WriteString(char)
+		}
+
+		if currentLineRuneCount < lineLength {
+			continue // next line
```
atonks2:
```suggestion
			continue // next rune
```

atonks2 (dismissed):

darwinz (approved):

reader.go
adamdecaf: thanks

atonks2 (approved):

PR #1398, opened 2024-03-19
Update openapi.yaml - addenda17/addenda18/format for entryHash

venukb:
https://github.com/moov-io/ach/blob/master/iatEntryDetail.go#L122-L128

`addenda17` and `addenda18` in `IATEntryDetail` is array Updating openapi spec to match this

-----
`entryHash` by default generates as `Integer` and entryHash values might be bigger and long is required.


adamdecaf (approved):

adamdecaf:
Thank you!

PR #1411, opened 2024-04-25
Add R62 to list of dishonored return codes

ckbaum:
In the NACHA rulebook Part 4.2: Table of Return Reason Codes, `R62` is included under "Codes to be Used by the ODFI for Dishonored Return Entries".


adamdecaf (approved):

adamdecaf:
Fixed in https://github.com/moov-io/ach/releases/tag/v1.37.5 - Appreciate the bug report!

PR #1433, opened 2024-05-31
[line-ending] allow specifying line ending via API

wokkaflokka:
Ability to specify line-ending is supported if accessing directly via Go. This change exposes ability to specify line-ending for API requests when formatting a file as plaintext.

This PR may require additional updates, opening to gain insight on the steps to move forward.

### DIFF
--- a/server/files.go
+++ b/server/files.go
@@ -351,6 +351,8 @@ type getFileContentsRequest struct {
 	ID string
 
 	requestID string
+
+	opts *ach.WriteOpts
 }
 
 type getFileContentsResponse struct {
@@ -366,7 +368,7 @@ func getFileContentsEndpoint(s Service, logger log.Logger) endpoint.Endpoint {
 			return getFileContentsResponse{Err: ErrFoundABug}, ErrFoundABug
 		}
 
-		r, err := s.GetFileContents(req.ID)
+		r, err := s.GetFileContents(req.ID, req.opts)
 
 		if logger != nil {
 			logger := logger.With(log.Fields{
--- a/server/files.go
+++ b/server/files.go
@@ -404,9 +404,14 @@ func decodeGetFileContentsRequest(_ context.Context, r *http.Request) (interface
 	}, nil
 }
 
-// GetLineEnding returns the header value for Line Endings
+// Inspects the `X-Line-Ending` header. If it is a valid value (CRLF | LF), returns the line ending associated
+// with the selected enum value. Otherwise, defaults to Unix-style newline characters.
 func GetLineEnding(r *http.Request) string {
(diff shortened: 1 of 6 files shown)


server/files.go
```diff
@@ -366,7 +368,7 @@ func getFileContentsEndpoint(s Service, logger log.Logger) endpoint.Endpoint {
 			return getFileContentsResponse{Err: ErrFoundABug}, ErrFoundABug
 		}
 
-		r, err := s.GetFileContents(req.ID)
+		r, err := s.GetFileContents(req.ID, req.opts)
```
adamdecaf: Don't we need to read LineEnding from the request? From a query param or Accept header?

wokkaflokka: @adamdecaf yes, we would need to take it from input. Either a param or a header could work. Right now, I think I'm expecting it as a JSON field -- `opts.lineEnding`

I was under the impression this option was being parsed automatically from the body, but I could very well be wrong. I can add a test to try and confirm.

writer.go
```diff
@@ -35,11 +35,26 @@ type Writer struct {
...
+}
+
 // NewWriter returns a new Writer that writes to w.
 func NewWriter(w io.Writer) *Writer {
+	return NewWriterWithOpts(w, nil)
+}
+
+// NewWriter returns a new Writer that writes to w.
+func NewWriterWithOpts(w io.Writer, opts *WriteOpts) *Writer {
+	lineEnding := "\n"
```
wokkaflokka, line 51: We could try to infer the default from the running operating system here, e.g. look at `runtime.GOOS`, if desired.

server/files.go
wokkaflokka: @adamdecaf thanks, I revisited this and understand your point now.

I added an HTTP header to convey this information and added tests verifying the behavior.

_Related to this -- what is the process for propagating changes from the openapi spec here to the generated client libraries?_

adamdecaf: I'm wondering if we should accept "CRLF" as an preference over the header accepting `\r` / `\n` as they may get wiped out by HTTP servers.

wokkaflokka: @adamdecaf I am inclined to agree.

I think an enum-like value accepting `CRLF | LF` would be a good choice.

Updated here: https://github.com/moov-io/ach/pull/1433/commits/044bd776fef0cff5e62ebfb4c80fb30c918cf68f

adamdecaf: Thanks

writer.go
adamdecaf: We've avoided that because it's hard to know what an HTTP server should return. If the server and client are on different OS's then who wins?

adamdecaf (approved):

PR #1439, opened 2024-06-16
Updated reference in comment to mixed credits and debits value. Adjusted small typo, capitlization.

Klasens:
I've updated one of the comments for Service Class Code value for mixed credits and debits. It's correctly labeled on the comment preceeding it, so I'm just changing it to reflect the 200 value above.


adamdecaf (approved):
Thanks

PR #1442, opened 2024-06-30
Update CONTRIBUTING.md (fixed bad links, minor spelling and grammar fixes)

n2847:
Made some small changes to ACH's CONTRIBUTING.md file, including fixing broken links:

- Fixed the “AUTHORS” link. The current link results in a 404 since the file is called “AUTHORS”, not “AUTHORS.md”.

- Updated the old “Go Code Review Comments” link so that the user isn’t required to manually redirect their browser.

- Fixed the numbering of instructions for adding SEC types (it should start with the number "1").

- Updated the "readACH" link, which currently gives a 404.

- Updated the "writeACH" link, which currently gives a 404.

- Updated several "Format Specification" links.

- Fixed some misspellings (for example, Cheny->Cheney, buide->guide, Wikipeda->Wikipedia)

- Changed the “I” in the word “FIle” from uppercase to lowercase.

- Changed one instance of “Nacha” to “NACHA” for consistency and correctness.

- Changed the scheme in the Wikipedia link from “http” to “https” in order to avoid an automatic redirect.

While the aforementioned changes should be good, there are still some questions that I have:
- The CONTRIBUTING.md file mentions a google group. Does that group still exist?
- It doesn’t look like anything has happened with the "Travis-CI" pipeline in quite some time (latest run was 4 years ago). Should references to it be removed?
- I couldn't find an updated PNC ACH File Specification. What link should be used?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,22 +1,22 @@
 # Contributing
 
-Wow, we really appreciate that you even looked at this section! We are trying to make the worlds best atomic building blocks for financial services that accelerate innovation in banking and we need your help!
+Wow, we really appreciate that you even looked at this section! We are trying to make the world's best atomic building blocks for financial services that accelerate innovation in banking and we need your help!
 
 You only have a fresh set of eyes once! The easiest way to contribute is to give feedback on the documentation that you are reading right now. This can be as simple as sending a message to our Google Group with your feedback or updating the markdown in this documentation and issuing a pull request.
 
 Stability is the hallmark of any good software. If you find an edge case that isn't handled please open an GitHub issue with the example data so that we can make our software more robust for everyone. We also welcome pull requests if you want to get your hands dirty.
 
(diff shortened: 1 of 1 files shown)


n2847:
In addition to the other questions mentioned in the pull request, I wondered about the "-bench=BenchmarkWEBDebitRead" Benchmarks example. That particular "go test" call doesn't reference an executable. Should it?

Here are the other questions that were included in the merge request:
- The CONTRIBUTING.md file mentions a google group. Does that group still exist?

- It doesn’t look like anything has happened with the "Travis-CI" pipeline in quite some time (latest run was 4 years ago). Should references to it be removed?

- I couldn't find a reference to an updated PNC ACH File Specification. Any idea where one might be?

@adamdecaf

adamdecaf:
Thanks for all the careful attention here. You're right in that we have some old information in that file. I'll clear it out.

PR #1477, opened 2024-08-28
Expose ValidateOpts field in File and FileHeader

kevinbarbour:
This is necessary so that `ValidateOpts` can be retained when files are submitted to the achgateway via the HTTP API.

Currently [it is documented](https://moov-io.github.io/achgateway/api/#post-/shards/-shardKey-/files/-fileID-) that the JSON payload can contain `ValidateOpts` but they aren't actually unmarshalled from the JSON as they are private in the structs and lacking JSON annotation.

### DIFF
--- a/file.go
+++ b/file.go
@@ -71,7 +71,7 @@ type File struct {
 	// ReturnEntries is a slice of references to file.Batches that contain return entries
 	ReturnEntries []Batcher `json:"ReturnEntries"`
 
-	validateOpts *ValidateOpts
+	ValidateOpts *ValidateOpts `json:"validateOpts,omitempty"`
 }
 
 // NewFile constructs a file template.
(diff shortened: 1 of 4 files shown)


adamdecaf:
This sounds like a bug in achgateway rather than in moov-io/ach

kevinbarbour:
Does the achgateway not use the models from moov-io/ach for its APIs?

adamdecaf:
What about https://github.com/moov-io/ach/pull/1481 as a fix?

kevinbarbour:
Thanks for the fix Adam, it looks like this should do what we need.

kevinbarbour:
I think the one thing I don't like about that approach is that it means if you are building ACH files in another service using the ach library and then submitting them to the achgateway, you can't simply marshal a `File` to JSON and retain the `ValidateOpts`. 

This way it requires a manual addition of the `ValidateOpts` to the JSON, and it doesn't get them in the OpenAPI spec for the achgateway if you are using a generated client.

adamdecaf:
> I think the one thing I don't like about that approach is that it means if you are building ACH files in another service using the ach library and then submitting them to the achgateway, you can't simply marshal a `File` to JSON and retain the `ValidateOpts`.

I can marshal `ach.File` and [see it retains ValidateOpts](https://github.com/moov-io/ach/pull/1481/commits/4a90c49441a7cc0b18d16278adba71e6b463fd52)


> This way it requires a manual addition of the `ValidateOpts` to the JSON, and it doesn't get them in the OpenAPI spec for the achgateway if you are using a generated client.

Ah yea, ValidateOpts is [missing from the OpenAPI spec](https://github.com/moov-io/ach/pull/1481/commits/24e9490df967a856c415fca1b2365bac5436f1b0) but that's fixable.

kevinbarbour:
Oh great - I guess I misunderstood the `json.Marshal` implementation, I didn't think it would retain it as a private field in the struct.

Thanks a lot for the quick fixes here!

adamdecaf:
`MarshalJSON` is implemented which lets you override the default behavior. We apply a trick to include ValidateOpts without having to expose it. Probably overkill now as it'd be more obvious with an exported ValidateOpts field. 

See: https://github.com/moov-io/ach/blob/master/file.go#L217

adamdecaf:
Are you good if I close this PR and merge/release https://github.com/moov-io/ach/pull/1481 ?

kevinbarbour:
Yep, all good with me. Thanks again!

PR #1511, opened 2024-11-21
Add DeleteEntry and DeleteADVEntry to Batcher

### LINKED ISSUE #1509: Is there a way to delete an entry from the batch?
### ACH Version

1.40.3

I'm looking for a way to delete an entry that has been added to the batch. 

**For example:** When I'm adding an entry to the batch during the NACHA file creation process, there are some situations where a particular entry causes the batch validation to fail. In those cases I want to remove that entry from the batch and move forward with adding that batch to the file. Upon scanning the `batcher` APIs, I couldn't find an API to delete an entry from the batch. So I was wondering if there is any alternate way to do the same thing?

Ability to delete an entry from the batch

### What did you see?

No API provided in `batcher`

By adding an entry to the batch and try removing it

kevinbarbour:
closes #1509 

Was also contemplating if it would be more useful to implement a `DeleteFunc` as the `slices` library provides, then it's easier to delete from the `Entries` slice based on arbitrary criteria instead of by maintaining or retrieving the pointer you want to delete. If that's preferable I can implement that.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 	"unicode/utf8"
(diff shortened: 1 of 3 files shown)


adamdecaf (approved):

adamdecaf:
Thinking about this more implementing them as DeleteFunc makes more sense. That lets callers control what is matched on to delete (and allows for deleting more than one entry)

kevinbarbour:
Sounds good - yeah that's how I was thinking. I will update it to add that.

adamdecaf (approved):

PR #1516, opened 2024-12-04
fix: enforce limits on max amount in entries, batches, files and merge

adamdecaf:


batchControl.go
```diff
@@ -147,6 +148,11 @@ func (bc *BatchControl) String() string {
 	return buf.String()
 }
 
+const (
+	// NachaBatchDebitCreditLimit is the maximum amount allowed by the Nacha format for a batch's debit/credit total (12 digits)
+	NachaBatchDebitCreditLimit = 9_999_999_999_99
```
darwinz, line 153: Shouldn't it be $99.999M?  This looks like $9.999B - which is two too many 9s

adamdecaf: The Amount fields in Ach are integers with two places for the decimal portion. This is the batch control which [has a field size of 12 digits](https://github.com/moov-io/ach/blob/master/batchControl.go#L195).

darwinz (approved):

PR #1518, opened 2024-12-04
fix: preserve FileIDModifier when flattening files

Klasens:
When flattening a file using the `/files/{fileID}/flatten` endpoint, the FileIDModifier value from the original file header was not being copied to the new flattened file, causing it to default back to "A". This fix preserves the original FileIDModifier when flattening files.


adamdecaf:
Thanks! Do you have a test that verifies this?

adamdecaf (approved):

Klasens:
Ahh right. I've added a test to file_test.go. I'm newer to Go so please let me know if something isn't looking right or idiomatic there.

PR #1521, opened 2024-12-05
Fix bug in Addenda02: Corrected ReferenceInformationTwoField to use ReferenceInformationTwo

mohdjishin:
Fix bug in Addenda02: Corrected ReferenceInformationTwoField to use ReferenceInformationTwo

In the **Addenda02** struct, **the ReferenceInformationTwoField method was incorrectly referencing ReferenceInformationOne instead of ReferenceInformationTwo**. This commit fixes the method to properly reference the correct field (ReferenceInformationTwo). This ensures that the correct data is returned and resolves any issues caused by the incorrect reference.


adamdecaf (approved):

adamdecaf:
Thanks! Released in https://github.com/moov-io/ach/releases/tag/v1.44.2

PR #1541, opened 2025-01-29
fix typos in batch code constants

bendrucker:
Fixes typos in the batch code constants' GoDocs. I happened to notice the typos in `PPD` while reading that description and ran a spellchecker on the rest while I was there.


adamdecaf (approved):
Thanks!

PR #1575, opened 2025-04-04
Store line numbers on each record

ckbaum:
Store line numbers on each record structure while reading a file, because:
- It can be tricky to correlate moov-parsed records to their record in the original file; when there's a non-format issue with a file provided by a client, we want to be able to clearly point them to the problematic entry.
- We've been seeing real-world cases where incoming files from the fed contain two entries with identical trace numbers and batch numbers. Access to the line number of the entries would allow us to unambiguously identify and dedup entries in this scenario.


adamdecaf (approved):

adamdecaf:
Yea, that's fine to add. I'll clarify these are not settable

Issue #1564, opened 2025-03-07
ACH Reader for UTF-8 File


hshih-lead:
### ACH Version

1.41.1

### What were you trying to do?

We're using the ACH `Reader` class to parse ACH files. Most of the time, files we parse from the Feds are encoded in ASCII text files, which works fine. Occasionally, we receive files encoded in UTF-8. This is likely due to names that are not in the standard ASCII range (hex `00` - `7F`). For example, `John DèS`.

### What did you expect to see?

Since `è` is in the extended ASCII range (valid for NACHA) , we expect it to be parsed properly by the Reader.

### What did you see?

However, we're seeing that `ach.NewReader()` seems to be interpreting that string in ISO 8859-1 encoding. So instead of seeing `è` in the reader, we're seeing `Ã¨`. In UTF-8, `è` has the byte value of `C3 A8`, which corresponds to `Ã` and `¨` in ISO 8859-1 respectively.

### How can we reproduce the problem?

1. Create a dummy ACH file in UTF-8 encoding, containing special character that is in the extended ascii range. For example, `è`.
2. Read the file and pass to ach.NewReader()
3. Store the result from achReader.Read() and print out the parsed results

adamdecaf:
We're using Go's official [`golang.org/x/net/html/charset`](https://pkg.go.dev/golang.org/x/net/html/charset) package for detecting windows 1252 / ISO 8859-1 encoding.  It looks like that's your issue (either incorrect detection of windows-1252 or imporper handling within that encoding).

windows-1252 was added quite a while ago https://github.com/moov-io/ach/pull/1272

I noticed you're on quite an old version of the library. Does this correctly parse on [v1.45.5](https://github.com/moov-io/ach/releases/tag/v1.45.5)?

hshih-lead:
Actually I realized that the line can be parsed properly if I **first encode the UTF8 data INTO Windows-1252** format before passing to ach reader. Not sure if this is an actual bug or a known behavior, that passing UTF8 data with special character into the reader (which calls charset) turns the data into something weird.

hshih-lead:
I wonder if this is a charset problem - that passing UTF8 content with characters in the extended ascii range (128-255) into `charset.NewReader(r, "text/plain")` [here](https://github.com/moov-io/ach/blob/master/reader.go#L149) makes bad encoding

So now I'm doing UTF8 (original content) -> Windows1252 (extra conversion) -> UTF8 (charset). It also works, just seem a bit odd.

[...]

PR #1582, opened 2025-04-10
New Validation Flag for Individual Name

hunshih:
Sometimes ODFI sends us entries with Individual Names that contain special characters outside of the alphanumeric range defined by Nacha. The new validation opt AllowSpecialCharactersInIndividualName will allow us to skip that check, and it's default to false for backward compatibility.

More discussion in https://github.com/moov-io/ach/issues/1564


adamdecaf:
Thanks for the PR. https://github.com/moov-io/ach/pull/1583 was also created for this which is a more encompassing solution.

PR #1583, opened 2025-04-10
Add AllowSpecialCharacters as a validation option

ckbaum:
Though the NACHA rulebook defines the set of characters that are considered "alphanumeric", in practice we sometimes receive incoming files with obscure special characters that we'd like to be able to accept. These are especially common in IAT entries.

Add a new validation option, `AllowSpecialCharacters`, which disables all `isAlphanumeric` checks. This is a more far-reaching version of the validation option added by https://github.com/moov-io/ach/pull/1582/files, which restricts the special character allowance to the IndividualName field (that's a little safer but sometimes insufficient)

### DIFF
--- a/addenda02.go
+++ b/addenda02.go
@@ -65,6 +65,8 @@ type Addenda02 struct {
 	validator
 	// converters is composed for ACH to GoLang Converters
 	converters
+	// validateOpts defines optional overrides for record validation
+	validateOpts *ValidateOpts
 }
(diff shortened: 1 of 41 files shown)


adamdecaf (approved):

adamdecaf:
Thanks. I think this will work for https://github.com/moov-io/ach/pull/1582 as well.

PR #1585, opened 2025-04-21
Return ParseErrors consistently when reading files

cjao-lead:
When `Reader.parseLine()` fails, it sometimes returns a `ParseError` and at other times returns a raw error like `ErrFileAddendaOutsideBatch` with no line context. This PR makes `parseLine()` always return a `ParseError` so that the error messages always include the number of the offending line.


adamdecaf (approved):

adamdecaf:
Thanks, this is defiantly an oversight.

PR #1623, opened 2025-07-01
Added a fix to remove trailing spaces in file control reserved

enikab:
This PR is based on a bug that was identified in the follow slack thread: https://moov-io.slack.com/archives/CD9J8EJKX/p1751301702060649?thread_ts=1749150613.975629&cid=CD9J8EJKX


reader.go
```diff
@@ -271,6 +271,8 @@ func (r *Reader) Read() (File, error) {
 		}
 	}
 
+	// Remove any trailing spaces within the file control reserved
+	r.File.Control.Reserved = strings.TrimSpace(r.File.Control.Reserved)
```
adamdecaf: We need to look at .validateOpts.PreserveSpaces before trimming

enikab: Done

adamdecaf (approved):

PR #1634, opened 2025-07-31
add arm64 build

tipputhynhean:
- Adding arm64 build

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,9 +74,16 @@ jobs:
       run: make dist-webui
 
   docker:
-    name: Docker
+    name: Docker ${{ matrix.arch }}
     needs: [testing, release]
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - runner: ubuntu-latest
+            arch: amd64
+          - runner: ubuntu-24.04-arm
+            arch: arm64
+    runs-on: ${{ matrix.runner }}
     steps:
     - name: Set up Go 1.x
       uses: actions/setup-go@v6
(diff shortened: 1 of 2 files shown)


adamdecaf:
Can we use Github's ARM runners instead?

tipputhynhean:
@adamdecaf Updated, please check. I'm not sure if I can clean up the docker and docker-hub in `Makefile` since I see it being used by `go.yml`

.github/workflows/release.yml
```diff
@@ -191,8 +191,20 @@ jobs:
...
-    - name: Docker
-      run: make docker
+    - name: Extract VERSION from Makefile
+      id: get_version
+      run: |
+        VERSION=$(make --no-print-directory get-version)
+        echo "version=$VERSION" >> $GITHUB_OUTPUT
+
+    - name: Build ARM64 and AMD64 Docker images
+      uses: docker/build-push-action@v5
```
adamdecaf: Can't we use an arm runner like `ubuntu-24.04-arm` to push architecture specific tags and combine them under a docker manifest? We do that internally a lot. Finding free arm runners (for OSS) is hard, but more are coming out. 

https://github.com/actions/partner-runner-images?tab=readme-ov-file#available-images

tipputhynhean: thank you @adamdecaf , I have updated it now, please advice if I need to change anything else

adamdecaf:
Can we leverage the makefile to run docker commands? That is a lot cleaner and we already read VERSION, so we don't need to duplicate that logic. The existing task can append the arch suffix, which can come from the action step if it's easier.

PR #1646, opened 2025-08-14
feat: Add v2 endpoint for structured file creation errors

viblo-majority:
This commit introduces a new `/v2/files/create` endpoint that returns structured JSON errors upon validation failure. The existing `/files/create` endpoint is unchanged.

The new endpoint provides a structured error response, which allows clients to programmatically handle errors without parsing strings.


viblo-majority:
Sorry, intendent to make a PR on my fork and not to here (until I have reviewed it a bit more)

Issue #1633, opened 2025-07-22
File Level Validation Option


hshih-lead:
### ACH Version

1.49.1

### What were you trying to do?

When performing incoming Fed NACHA file, we actually only care about "file level" validation, so things like File Header, Control, Addenda Count, etc.
However, seems like the current [validation call](https://github.com/moov-io/ach/blob/master/file.go#L655) also does entry level validation here: https://github.com/moov-io/ach/blob/master/file.go#L831

Currently we have two options:
(1) Use the `SkipAll` option: This skips both the file level validation as well as entry level, or
(2) Specify all the options listed in the struct [ValidateOpts](https://github.com/moov-io/ach/blob/master/file.go#L679), though some entries type might not check those flags.

**Is it possible to add a new flag called `SkipBatchValidation` that skips [this batch loop](https://github.com/moov-io/ach/blob/master/file.go#L831) which performs entry level validation?**

### What did you expect to see?

When we specify with the new flag, only file level info gets validated.

adamdecaf:
These broader options are tricky to get right in the code. Speaking of are you wanting to skip batch level validations (and entries) or only validation on EntryDetail records? 

The link for "this batch loop" points towards batch validation, but the name `SkipEntryLevelValidation` would apply to EntryDetails.

hshih-lead:
Sorry to confuse! I should have used more precise language. Actually **skipping Batch level** validation (which includes EntryDetails) will suffice for us. So the flag would just be something like `SkipBatchValidation`

adamdecaf:
If you can open a PR I can help review it and get it merged. We'll need to double check the batch validations are checked properly with this flag.

adamdecaf:
This will be included in the next release.

PR #1648, opened 2025-08-19
feat: add BypassBatchValidation to skip batch validation

hunshih:
https://github.com/moov-io/ach/issues/1633
- Added `BypassBatchDetailsValidation` to skip batch level validation
- Added unit test

### DIFF
--- a/file.go
+++ b/file.go
@@ -755,6 +755,9 @@ type ValidateOpts struct {
 	// AllowEmptyIndividualName will skip verifying IndividualName fields are populated
 	// for SEC codes that require the field to be non-blank (and non-zero)
 	AllowEmptyIndividualName bool `json:"allowEmptyIndividualName"`
+
+	// BypassBatchValidation will skip validation for batches in a file and only validate file header and control info
+	BypassBatchValidation bool `json:"bypassBatchValidation"`
 }
 
 // merge will combine two ValidateOpts structs and keep any non-zero field values.
@@ -787,6 +790,7 @@ func (v *ValidateOpts) merge(other *ValidateOpts) *ValidateOpts {
 		AllowZeroEntryAmount:             v.AllowZeroEntryAmount || other.AllowZeroEntryAmount,
 		AllowSpecialCharacters:           v.AllowSpecialCharacters || other.AllowSpecialCharacters,
 		AllowEmptyIndividualName:         v.AllowEmptyIndividualName || other.AllowEmptyIndividualName,
+		BypassBatchValidation:            v.BypassBatchValidation || other.BypassBatchValidation,
 	}
 
 	if v.CheckTransactionCode != nil {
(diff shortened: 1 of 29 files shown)


file.go
```diff
@@ -755,6 +755,10 @@ type ValidateOpts struct {
 	// AllowEmptyIndividualName will skip verifying IndividualName fields are populated
 	// for SEC codes that require the field to be non-blank (and non-zero)
 	AllowEmptyIndividualName bool `json:"allowEmptyIndividualName"`
+
+	// BypassBatchDetailsValidation will skip validation for individual batches in a file
+	// and only validate file header and control info
+	BypassBatchDetailsValidation bool `json:"bypassBatchValidation"`
```
adamdecaf: Can we name this `BypassBatchValidation`?

file.go
```diff
@@ -828,9 +832,11 @@ func (f *File) ValidateWith(opts *ValidateOpts) error {
 			return NewErrFileCalculatedControlEquality("BatchCount", len(f.Batches), f.Control.BatchCount)
 		}
 
-		for _, b := range f.Batches {
-			if err := b.Validate(); err != nil {
-				return err
+		if !opts.BypassBatchDetailsValidation {
+			for _, b := range f.Batches {
+				if err := b.Validate(); err != nil {
```
adamdecaf, line 837: We need to check inside each `func (b *batch) build()` if we should skip validations too. 

Does this option apply to IAT batches? By the name it would skip validations for IAT batches as well.

file.go
hshih-lead: Oops good call. Updated

file.go
hunshih: Added flag to skip validation in IAT batches as well.
Also I wonder if I can just check for `BypassBatchDetailsValidation` wherever we see `SkipAll`. The only difference for those two flags would be in `func (f *File) ValidateWith(opts *ValidateOpts)`, where I actually want to run some file level checks. Does that work?

hunshih: Let me know if I should still add `BypassBatchDetailsValidation` in `func (b *batch) build()`, since that's used in some cases to create batches and files.

adamdecaf: I think that will work.

reader.go
```diff
@@ -1000,7 +1000,7 @@ type canValidate interface {
 }
 
 func maybeValidate(rec canValidate, opts *ValidateOpts) error {
-	if opts != nil && opts.SkipAll {
+	if opts != nil && (opts.SkipAll || opts.BypassBatchValidation) {
```
adamdecaf: This isn't correct. Are we sure `maybeValidate` is only called on batch objects?

adamdecaf:
This is missing support in the HTTP server (under `./server`)

hunshih:
> This is missing support in the HTTP server (under `./server`)

Added

reader.go
hunshih: Good catch. `maybeValidate ` seems to be called on File Header/Control, Batch objects, Entry Details, and Addendas.
I added some logic such that if `BypassBatchValidation` is specified, validation should ONLY be run for file type objects. Return nil for batch, entry, and addendas.

adamdecaf:
Thanks! I've cleaned up the commits and removed the public API that was added.

PR #1651, opened 2025-08-25
fix: add content-specific endpoints for Java clients

### LINKED ISSUE #1650: Create file & segment endpoints broken for Java clients
### ACH Version

v1.48.0

The `POST /files/{fileID}` endpoint officially allows callers to create an ACH file from either a JSON schema or a raw ACH file. The endpoint determines which code path to use from the content type header in the request. My java app needs to be able to leverage both of these code paths in order to create new ACH files (via JSON) and to work with existing ACH files (via raw text).

I expected the OpenAPI generated Java client library to have support for using all content types defined by the `POST /files/{fileID}` endpoint.

### What did you see?

Unfortunately, there is a [bug](https://github.com/OpenAPITools/openapi-generator/issues/17877) in the Java Open API generator (reported over a year ago) that prevents library callers from being able to use multiple content type definitions defined by the same API endpoint. Practically, this means that Java clients using the ACH create file endpoint cannot access the code path to create a file from raw ACH text. Our app therefore cannot make use of both the raw ACH file creation and json ACH file creation features.

Note: based on the spec, the same issue will also exist for the `/segment` api endpoint.

Use the latest open api generator to create a java client library using the openapi spec defined in this repo. Attempt to use the generated library to create an ACH file from raw text. It is not possible. The only workaround is to comment out the `application/json` part of the spec when generating the client library, but then the JSON code path is unusable.

eliana404:
Fixes https://github.com/moov-io/ach/issues/1650

I added content-type-specific endpoints for both the file creation and file segmentation functionality, since those endpoints were the only two that I saw which defined multiple content types.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -270,6 +270,316 @@ paths:
           description: Permanently deleted File.
         '404':
           description: A File with the specified ID was not found.
+  /files/{fileID}/json:
+    post:
+      tags: ['ACH Files']
+      summary: Create File
+      description: Create a new File object from the JSON representation.
+      operationId: createFile
+      parameters:
+        - name: X-Request-ID
+          in: header
+          description: Optional Request ID allows application developer to trace requests through the system's logs
+          example: "rs4f9915"
+          schema:
+            type: string
+        - name: fileID
+          in: path
+          description: File ID
+          required: true
+          schema:
+            type: string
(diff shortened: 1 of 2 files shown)


eliana404:
Closing in favor of https://github.com/moov-io/ach/pull/1652

PR #1653, opened 2025-08-27
server: expose file create endpoint under agnostic request body

adamdecaf:
Issue: https://github.com/moov-io/ach/issues/1650

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -61,12 +61,7 @@ paths:
       description: Create a new File object from either the plaintext or JSON representation.
       operationId: createFile
       parameters:
-        - name: X-Request-ID
-          in: header
-          description: Optional Request ID allows application developer to trace requests through the system's logs
-          example: "rs4f9915"
-          schema:
-            type: string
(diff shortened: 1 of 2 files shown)


openapi.yaml
```diff
@@ -270,6 +182,68 @@ paths:
           description: Permanently deleted File.
         '404':
           description: A File with the specified ID was not found.
+  /files/{fileID}/new:
+    post:
+      tags: ['ACH Files']
+      summary: Create New File
+      description: Create a new File object from either the plaintext or JSON representation.
```
eliana404: You might want to clarify that both are plaintext (string) and one is NACHA while the other is a JSON string. Something that helps distinguish this endpoint from the existing one.

adamdecaf: Both endpoints are the same behind the scenes and can accept Nacha or JSON formatted files. 

I can mention the new endpoint is helpful for generated clients that can't handle the older endpoint.

eliana404 (approved):

PR #1655, opened 2025-08-28
fix: validate entry AddendaRecordIndicator

### LINKED ISSUE #1647: Batch Validation Missed Addenda Format
### ACH Version

1.45.5

We rely on the batch validation to make sure we're structuring our NACHA file with the correct format, using this validation call:
https://github.com/moov-io/ach/blob/master/batcher.go#L46

We expected that call to validate the batch format catching all errors. More specifically, the `addendaRecordIndicator` and the number of addendas in the entry should match.

### What did you see?

We accidentally set the `addendaRecordIndicator` field as 1 in the entry detail without actually having addendas set. This bad format led to entries being rejected by the Fed via return code R25. 

Set the addendaRecordIndicator to a value that does not match the actual number of addendas in the entry. The batch validation will not catch that as an error.

adamdecaf:
Fixes: https://github.com/moov-io/ach/issues/1647

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -413,6 +413,11 @@ func (batch *Batch) build() error {
 		for i, entry := range batch.Entries {
 			entryCount += 1 + entry.addendaCount()
 
+			// addendaCount := entry.addendaCount()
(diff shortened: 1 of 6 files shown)


adamdecaf:
Superseded by https://github.com/moov-io/ach/pull/1864

🤖 grok

PR #1688, opened 2025-10-31
feat: add SkipFileCreationValidation 

### LINKED ISSUE #1681: `fileCreationDate` syntax/length not checked
### ACH Version

v1.53.0

We use the API to `/create` ACH files from a json structure, and use `/content` to get the raw file.

We called the create endpoint with an ACH file that had `2510081332` in `fileHeader.fileCreationDate`. and it didn't raise an error, even if the format is wrong. Expected is `YYMMDD`, which is correctly documented in the openapi spec.

The resulting raw file has strange values at the date places (000000 in the second row (520) and the first row (101) missed 6 characters!).

I'm sure I can provide a sample file but it's very easy, simply use `2510081332` in `fileHeader.fileCreationDate` (`"fileCreationTime": ""`).

An error telling me the date format is incorrect.

### What did you see?

A broken raw file with wrong/missing values and wrong line lengths.

Use the ACH API (Endpoint `/create`) with a json that has `2510081332` in `fileHeader.fileCreationDate` (`"fileCreationTime": ""`), then use `/content` to generate a broken raw file.

adamdecaf:
Fixes https://github.com/moov-io/ach/issues/1681

### DIFF
--- a/docs/create-file.md
+++ b/docs/create-file.md
@@ -62,6 +62,7 @@ Example: `POST /files/create?requireABAOrigin=true&bypassDestination=true`
 | `preserveSpaces`                   | `PreserveSpaces`                   |
 | `requireABAOrigin`                 | `RequireABAOrigin`                 |
 | `skipAll`                          | `SkipAll`                          |
+| `skipFileCreationValidation`       | `SkipFileCreationValidation`       |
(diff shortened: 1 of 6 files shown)


wadearnold:
There are still some issues to be solved in this project!

darwinz (approved):

PR #1690, opened 2025-11-04
feat:  add TXP ToString method

darwinz:


ahrtmn (approved):

addenda/txp_test.go
```diff
@@ -409,3 +409,167 @@ func TestIsTXPFormat(t *testing.T) {
...
+			// Parse the serialized string again
+			reparsed, err := ParseTXP(serialized)
+			if err != nil {
+				t.Fatalf("failed to reparse serialized TXP: %v", err)
+			}
+
+			// Verify the round-trip results match
+			if reparsed.TaxIdentificationNumber != txp.TaxIdentificationNumber {
+				t.Errorf("TaxIdentificationNumber: expected %s, got %s", txp.TaxIdentificationNumber, reparsed.TaxIdentificationNumber)
+			}
```
adamdecaf: Can we use testify/require instead?

addenda/txp.go
```diff
@@ -196,6 +196,38 @@ func ParseTXP(paymentInfo string) (*TXP, error) {
 	return txp, nil
 }
 
+// ToString serializes the TXP object into a TXP-formatted string
+// The format matches the expected TXP addenda format:
+// TXP*tax_id*tax_type*date*type1*amount1*type2*amount2*type3*amount3*taxpayer_verification\
+func (txp *TXP) ToString() string {
```
adamdecaf: The name `String() string` would match this library better. We don't want String() and ToString() - that'd be confusing.

adamdecaf (approved):

PR #1724, opened 2026-01-01
feat: add v2 API with structured validation errors

### LINKED ISSUE #1625: Support structured validation errors in HTTP API
### ACH Version

1.49.1

I’m using the POST /files/create endpoint to validate incoming ACH files and then map any validation failures to the appropriate Rxx return codes. Right now, when validation fails the API returns one error string with newline-separated entries. That forces my client to split on \n and regex-parse each line to figure out which record failed and then categorize the return code per entry.

Is it possible to include a opt-in flag (in example, ?structuredErrors=true) on POST /files/create so that on failure the API returns a JSON array of error objects, something like:

```
...
  "errors": [
    {
      "lineNumber": 2,
      "recordType": "EntryDetail",
      "errorType":  "FieldError",
      "fieldName":  "PaymentTypeCode",
      "message":    "PaymentTypeCode is a required field"
    }
  ]
}
```

This would let my client unmarshal a structured list of validation failures and map each directly to an appropriate R-code (R26, R27, etc.), rather than depending on manual text parsing. Parsing the error string directly feels a bit error-prone and difficult to maintain if the error format ever changes. Is having a structure error object something you think could be useful?

MorganaFuture:
Add /v2/files/* and /v2/files/{id}/batches endpoints that return all validation errors in structured JSON format instead of a single error string.

Add ValidateAll() methods to File and Batch types to collect all errors. v1 API unchanged.

Closes #1625

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -25,6 +25,8 @@ import (
 	"strconv"
 	"strings"
 	"unicode/utf8"
+
+	"github.com/moov-io/base"
 )
 
 // Batch holds the Batch Header and Batch Control and all Entry Records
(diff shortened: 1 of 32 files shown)


adamdecaf:
@MorganaFuture if we're going to make a v2 of endpoints I'd really like to get off go-kit and use gorilla directly.

PR #1745, opened 2026-02-11
fix: propagate file validation options to batches during flatten

### LINKED ISSUE #1744: FlattenBatches() doesn't preserve batch validation options during Copy()
### ACH Version

1.55.3

  When FlattenBatches() consolidates batches, it creates new batches via Copy() but doesn't preserve the batch-level validation options. This causes validation to fail when using options like
  AllowEmptyIndividualName.

  Expected Behavior

  Batch validation options should be preserved when batches are copied during the flatten operation.

  Current Behavior

  1. Original batches have validation options set via batch.SetValidation()
  2. FlattenBatches() calls Copy() which creates new batches without validation options
  3. New batches fail validation when batch.Create() is called
  4. Errors are silently ignored at line 95 in file_flattener.go
  5. Result: no batches are added to the flattened file

Validation options propagated and new batches created in flattened file

### What did you see?

panic: batch #1 (WEB) FieldError IndividualName  is a mandatory field and has a default value, did you use the constructor? [recovered, repanicked]

```
// Create batch with validation option
  batch, _ := ach.NewBatch(batchHeader)
  batch.SetValidation(&ach.ValidateOpts{
      AllowEmptyIndividualName: true,
  })

  // Add entry with spaces for individual name
  entryDetail := ach.NewEntryDetail()
  entryDetail.IndividualName = "                      " // 22 spaces
  batch.AddEntry(entryDetail)
  batch.Create() // ✅ Works

  // Add to file
  file := ach.NewFile()
  file.AddBatch(batch)
  file.Create() // ✅ Works

  // Flatten fails because validation options not preserved
  flattenedFile, err := file.FlattenBatches() // ❌ Fails
  // Error: "must have []*Batches or []*IATBatches to be built
```

vkemthur-luna:
Fixes #1744

This change ensures that validation options set on a File (such as AllowEmptyIndividualName) are properly propagated to batches during the FlattenBatches operation.

Previously, when flattening batches, validation options were not inherited by the newly created batches, causing validation errors even when the file had appropriate options set.

Changes:
- Added error checking in Flatten() - errors from AddToFile() were being silently ignored
- Updated mergeableBatcher.AddToFile() to inherit validation options from file and propagate them to batches and entries
- Updated mergeableIATBatch.AddToFile() with same validation inheritance for IAT batches
- Added comprehensive unit tests for validation propagation

This allows users to set validation options once at the file level and have them automatically applied during batch flattening.

### DIFF
--- a/file_flattener.go
+++ b/file_flattener.go
@@ -92,7 +92,9 @@ func Flatten(originalFile *File) (*File, error) {
 	sort.Slice(allBatches, func(i int, j int) bool { return allBatches[i].GetBatchNumber() < allBatches[j].GetBatchNumber() })
 
 	for i := range allBatches {
-		allBatches[i].AddToFile(newFile)
+		if err := allBatches[i].AddToFile(newFile); err != nil {
+			return nil, err
+		}
 	}
 
 	if err := newFile.Create(); err != nil {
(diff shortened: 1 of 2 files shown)


adamdecaf (approved):

adamdecaf:
Thanks!

Issue #1570, opened 2025-03-25
Enable file/batch balancing validations but disable all others


palsri2004:
### ACH Version

Latest

### What were you trying to do?

Hi,

I have looked into custom validations that Moov provided and didn't find a way to enable/disable balancing validations which will check if file totals matches with batch totals and transaction totals. I would like to disable all validations but this.  Can you please provide the way how to enable amount balancing validations and disable rest others.

Thanks,

### What did you expect to see?

I want to enable file totals/batch totals validations as a toggle and disable all other validations in NACHA file.

### What did you see?

I see that batch total is out of balance if I enable any of the validation but when I skip all then the balancing validations is also disabled. I couldn't find a toggle by which I can enable/disable balancing validation.

### How can we reproduce the problem?

Try with a batch with mismatch amount with file and then skip all validations. The file will pass. However when you disable any custom validations that Moov provided, you will get an error that batch is out of balance.

palsri2004:
Hello, is there any update on this?

viggykuppu:
@adamdecaf Hi Adam, I'm interested in implementing this feature. Is this a feature that would be accepted? I know it's a bit because of the idea of skipping all validations except batch/file totals isn't quite straightforward, but we're looking for this feature in production. We could write our own logic on top of Moov to accomplish this, but we were thinking it might be generally useful. I'd love to pick your brain on this if you have some time as well.

adamdecaf:
We don't have a good pattern for "disable all; enable X, Y, Z" validations. That makes your initial use-case harder to support. We've tried to keep the library/industry pushing towards Nacha compliance. 

Initially if all you're worried about is entry/batch totals equaling file totals I would have you implement that yourself. It would be easy enough to read a file (or call .Create() on each batch, then file) and check the totals.

adamdecaf:
@viggykuppu what would your implantation look like? Can you give a high-level overview / pseudocode?

viggykuppu:
@adamdecaf I have a PR here that outlines an approach: https://github.com/moov-io/ach/pull/1748 Can you let me know if you have feedback?

PR #1748, opened 2026-02-18
feat: adds a new validation option to only check batch + file totals and entry counts

viggykuppu:
Adds functionality requested in https://github.com/moov-io/ach/issues/1570

This is performed using a new Validation Option `skipAllValidationsExceptTotals`.  This changes file and batch validation logic to only check the credit and debit entry amounts + the entry addenda counts.

### DIFF
--- a/batchACK.go
+++ b/batchACK.go
@@ -41,6 +41,17 @@ func (batch *BatchACK) Validate() error {
 	if batch.validateOpts != nil && (batch.validateOpts.SkipAll || batch.validateOpts.BypassBatchValidation) {
 		return nil
 	}
+	if batch.validateOpts.SkipAllValidationsExceptTotals {
+		err := batch.isBatchEntryCount()
+		if err != nil {
+			return err
+		}
+		err = batch.isBatchAmount()
+		if err != nil {
+			return err
+		}
+		return nil
+	}
 
 	// basic verification of the batch before we validate specific rules.
 	if err := batch.verify(); err != nil {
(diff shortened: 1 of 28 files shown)


adamdecaf:
This is a bit convoluted.. I'm not a fan of duplicating the validation logic for an "Skip All Except X and Z". It seems like you're trying to disable all validations, but then add a couple back into the list of what's ran.. That might be better served by a different API for validations.

viggykuppu:
@adamdecaf Yeah, I think that's a good point; I definitely felt like it was a bit clunky as I was implementing it. Are you essentially suggesting adding a new function like `ValidateOnlyTotals`?

adamdecaf:
Not `ValidateOnlyTotals`, but perhaps wrap the existing functions to calculate totals so you could `SkipAll: true` and call that new `ValidateTotals() error` func? 

I'd want the validations written once and called from a couple spots. 

Are you using the HTTP api? I think we could see what `?skipAll=true&validateTotals=true` would look like.

viggykuppu:
@adamdecaf Okay, so just to make sure I understand you correctly, you're suggesting to keep the new validation option, and just pull the total validation logic into a separate function?

viggykuppu:
@adamdecaf Oh, I see, you're suggesting to add a validation flag `validateTotals` that will run the totals even if `skipAll=true`?

adamdecaf:
I think adding a function that calculates the totals, which can be called when SkipAll=true

viggykuppu:
@adamdecaf What's your thought on how to expose the ability to check the totals when SkipAll=true? Should this be a validation option that only applies if SkipAll=true?

viggykuppu:
@adamdecaf Or are you saying this should be separated out from the existing validation logic entirely? So the change would just be exposing a new function ValidateTotals() that does the aforementioned total + entry/addenda count validations?

adamdecaf:
My bad for missing these - your [new PR](https://github.com/moov-io/ach/pull/1756) looks like a good start.  Thanks

PR #1809, opened 2026-07-11
perf: add ASCII record parsing fast paths

khancepts101:
# ASCII fast-path benchmark results

## Summary

`EntryDetail.Parse` and `Addenda05.Parse` now use fixed byte offsets for 94-byte ASCII records. Records with a different byte length or any non-ASCII byte use the preserved rune-safe parser.

- `EntryDetail`: approximately 2.1x faster in isolation, with no allocation change.
- `Addenda05`: approximately 7.1x faster in isolation and eliminates two allocations (82 bytes) per record.
- Addenda-heavy CCD+ reader workload: approximately 13% faster with about 200,000 fewer allocations and 8.2 MB less allocated per 100,000 entries.
- PPD reader workload: no full-pipeline speedup is claimed because the measured ranges overlap and variance exceeded 5%.

Add parity coverage for edge cases and real repository fixtures, isolated fast-vs-slow benchmarks with package-level sinks, and an addenda-heavy CCD+ variant of `Benchmark_ReadLargeFile` with one Addenda05 per entry.

## Environment and method

- Go 1.26.5
- `darwin/amd64`
- VirtualApple 2.50 GHz
- Isolated benchmarks used package-level sinks, `-benchmem`, and three repetitions.
- Full-pipeline samples include gzip decompression, parsing, object construction, and `File.Validate`.
- Full-pipeline files contain 200 batches and exactly 100,000 entries.
- The CCD+ shape contains one Addenda05 record per entry.

The rune-safe baseline was measured by temporarily routing both parsers directly to their preserved rune-safe implementations. The production fast paths were restored after baseline measurement.

## Isolated parsing

| Parser      | Rune-safe path    | ASCII path        | Speedup    | Allocations  |
|-------------|-------------------|-------------------|------------|--------------|
| EntryDetail | 185.9–192.9 ns/op | 90.28–90.37 ns/op | 2.06–2.14x | 0 → 0        |
| Addenda05   | 576.3–577.4 ns/op | 81.38–81.70 ns/op | 7.05–7.10x | 2 (82 B) → 0 |

EntryDetail retains field conversion and validation-option work on both paths: numeric conversion, whitespace handling, SEC-code comparisons, and field assignments. 

## Full reader pipeline

Five blocked samples were collected per implementation with `-benchtime=1x`:

| File shape                | Rune-safe parsers | ASCII fast paths | Mean change  | Allocations           |
|---------------------------|-------------------|------------------|--------------|-----------------------|
| PPD, no addenda           | 223.9–260.3 ms    | 216.1–240.8 ms   | Not claimed  | ~203,664 → ~203,664   |
| CCD+, Addenda05 per entry | 441.8–459.3 ms    | 385.8–393.7 ms   | 13.3% faster | ~1,105,800 → ~905,802 |

The PPD ranges overlap and varied by 11.0% for the ASCII path and 15.4% for the rune-safe path. The CCD+ ranges do not overlap and varied by 2.0% and 3.9%, respectively.

| Repetition | Rune-safe range | ASCII range    | Mean improvement |
|------------|-----------------|----------------|------------------|
| 1          | 428.8–440.4 ms  | 372.5–384.2 ms | 13.2%            |
| 2          | 428.8–437.7 ms  | 368.9–377.5 ms | 13.7%            |
| 3          | 432.2–441.1 ms  | 374.7–382.9 ms | 13.0%            |

The interleaved measurements support approximately 13% as the full-pipeline improvement for this addenda-heavy workload. They should not be extrapolated to other ACH file shapes.

## Correctness and scope

Parity tests compare complete parsed structs from the ASCII and rune-safe paths using:

- Empty and wrong-length records.
- Records containing non-ASCII bytes.
- Hand-written valid records.
- EntryDetail and Addenda05 records sampled from `test/testdata/*.ach`.

Wrong-length and non-ASCII records always take the preserved rune-safe path. `go test ./...` passes.


adamdecaf (approved):

adamdecaf:
Thanks. I hadn't benched the two against each other like this to see that keeping the ASCII-only path was that helpful, but 10-15% is an important enough benefit. Plus the format is stable so maintenance on the two paths is almost zero effort.

PR #1834, opened 2026-08-12
fix: Iterator scanner errors and fixed-width rune splitting

SashaMIT:
## Problem

Two robustness issues in the streaming and fixed-width read paths, both reachable through `NewIterator`:

**1. `Iterator.NextEntry` silently drops entries after any line over 64KB**

`bufio.Scanner` stops permanently once a line exceeds `bufio.MaxScanTokenSize` (64KB). `NextEntry` never checks `i.scanner.Err()` after its scan loops, so when this happens the iterator returns `(nil, nil, nil)`, the exact clean end-of-file signal. Every entry after an overlong line is silently discarded with success semantics. The sibling `Reader.Read` path already checks `scanner.Err()`.

**2. Fixed-width splitter desynchronizes on multibyte input and drops the trailing record**

`processFixedWidthFile` breaks records on byte offsets (`(i+1) % RecordLength`) while the record contract is 94 characters. When the first line contains multibyte UTF-8 characters, record boundaries shift and any leftover bytes after the last 94-byte boundary are discarded without an error.

## Fix

- `iterator.go`: check `i.scanner.Err()` after both scan loops and return it wrapped with line-number context.
- `reader.go`: split fixed-width input on rune boundaries instead of byte offsets, and return an error if a partial record remains after the last full record.

## Tests

- `TestIterator/scanner_error_on_overlong_line`: a file with a line over `bufio.MaxScanTokenSize` now returns an error containing "token too long" instead of a clean EOF. Fails without the fix.
- `TestIterator/fixed-width_file_with_multibyte_characters`: a fixed-width file whose single line contains a multibyte character (rune count still an exact multiple of 94) now parses all records cleanly. Fails without the fix.

`go build`, `go vet`, and `go test ./...` pass.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -109,6 +109,9 @@ start:
 				break
 			}
 		}
+		if err := i.scanner.Err(); err != nil {
+			return nil, nil, fmt.Errorf("scanning line %d failed: %w", i.reader.lineNum+1, err)
+		}
 		// If we've exhausted all lines in the reader then quit
 		if line == "" || allSpaces(line) {
 			return nil, nil, nil
@@ -181,6 +184,9 @@ start:
 						return bh, returnableEntry, nil
 					}
 				} else {
+					if err := i.scanner.Err(); err != nil {
+						return nil, nil, fmt.Errorf("scanning line %d failed: %w", i.reader.lineNum+1, err)
+					}
 					break // quit processing if we can't read another line
 				}
 			}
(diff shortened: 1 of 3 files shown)


adamdecaf (approved):

SashaMIT:
Thanks for the merge.

PR #1842, opened 2026-08-20
Let the debit-only SEC types validate after a reversal

karpovantonme:
`File.Reversal` turns debits into credits and sets `ServiceClassCode` to `CreditsOnly`, which is what a reversal is. Seven batch types then reject the result, because their validation only ever expected the forward direction: ARC, BOC, POP, RCK, XCK, TRC and TRX all refuse `CreditsOnly` in `Validate` and refuse a credit transaction code in `InvalidEntries`.

TEL is the only one that knows about reversals, and it learned in place: the comparison sits inline in `batchTEL.go:69` with its own `strings.EqualFold`. The other 7 came from the same template and were never taught.

`Reversal` returns nil in every case, so nothing says so at the time. The file is simply invalid from then on.

Reproduced by building a forward batch of each type, calling `Reversal`, and validating:

| SEC | before | after |
|---|---|---|
| ARC | `ServiceClassCode header SCC is not valid for this batch's type: 220` | passes |
| BOC | same | passes |
| POP | same | passes |
| XCK | same | passes |
| TRC | same | passes |
| TRX | same | passes |
| TEL | passes | passes |

The comparison TEL did inline is now `Batch.IsReversal`, and the 7 types call it in both places. Nacha writes `REVERSAL` left justified in a fixed-width field, so the helper trims and compares without case, exactly as the TEL code already did.

## RCK, which this does not fix

RCK requires its Company Entry Description to be `REDEPCHECK`, and `Reversal` overwrites that field with `REVERSAL`. The two rules cannot both hold, so an RCK batch can never validate after a reversal no matter what the transaction codes say. `Create` stays quiet about it and only `Validate` complains, which is later than a caller would want to hear.

Whether Nacha permits reversing an RCK entry at all is a question for you rather than something for me to guess at, so `TestReversal_RCKCannotValidate` records the behaviour as it stands instead of asserting what it should be. If reversing RCK is meant to be impossible, `Reversal` returning an error for it would say so at the right moment, and I am happy to send that as a separate change.

## Checks

71 packages pass. The new tests are `TestReversal_DebitOnlySECTypes`, `TestReversal_RCKCannotValidate` and `TestBatch_IsReversal`.

How I found it: I was looking for places where the same rule is written in several files and only one copy was updated. `REVERSAL` appears in exactly one batch type out of eight, and the other seven are near-identical to it.

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -290,19 +290,6 @@ func ConvertBatchType(b Batch) Batcher {
 //
 // Create implementations are free to modify computable fields in a file and should
 // call the Batch's Validate function at the end of their execution.
-// IsReversal reports whether this batch was turned into a reversal, which the
-// Nacha rules signal by putting "REVERSAL" in the Company Entry Description.
-//
-// A reversal moves funds the other way, so a batch type that normally carries
-// debits only carries credits here, and its ServiceClassCode becomes
-// CreditsOnly. Validation that assumes the forward direction has to know about
-// this or it rejects a file the library itself produced.
-func (batch *Batch) IsReversal() bool {
-	if batch.Header == nil {
-		return false
-	}
-	return strings.EqualFold(strings.TrimSpace(batch.Header.CompanyEntryDescription), "REVERSAL")
-}
 
 func (batch *Batch) Create() error {
 	return errors.New("use an implementation of batch or NewBatch")
@@ -1096,6 +1083,14 @@ func (batch *Batch) addendaFieldInclusionReturn(entry *EntryDetail) error {
 		return batch.Error("Addenda99", ErrFieldInclusion)
 	}
 	return nil
+}
+
+// IsReversal determines if a batch carries reversing entries - see File.Reversal
+func (batch *Batch) IsReversal() bool {
+	if batch.Header == nil {
+		return false
+	}
+	return strings.EqualFold(strings.TrimSpace(batch.Header.CompanyEntryDescription), "REVERSAL")
 }
 
 // IsADV determines if a batch is batch type ADV - BatchADV
--- a/batch.go
+++ b/batch.go
@@ -290,6 +290,20 @@ func ConvertBatchType(b Batch) Batcher {
 //
 // Create implementations are free to modify computable fields in a file and should
 // call the Batch's Validate function at the end of their execution.
+// IsReversal reports whether this batch was turned into a reversal, which the
+// Nacha rules signal by putting "REVERSAL" in the Company Entry Description.
+//
+// A reversal moves funds the other way, so a batch type that normally carries
+// debits only carries credits here, and its ServiceClassCode becomes
+// CreditsOnly. Validation that assumes the forward direction has to know about
+// this or it rejects a file the library itself produced.
+func (batch *Batch) IsReversal() bool {
+	if batch.Header == nil {
+		return false
+	}
+	return strings.EqualFold(strings.TrimSpace(batch.Header.CompanyEntryDescription), "REVERSAL")
+}
+
 func (batch *Batch) Create() error {
 	return errors.New("use an implementation of batch or NewBatch")
 }
--- a/batchARC.go
+++ b/batchARC.go
@@ -62,10 +62,14 @@ func (batch *BatchARC) Validate() error {
 		return batch.Error("StandardEntryClassCode", ErrBatchSECType, ARC)
 	}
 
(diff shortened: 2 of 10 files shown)


batch.go
```diff
@@ -290,6 +290,20 @@ func ConvertBatchType(b Batch) Batcher {
...
+// A reversal moves funds the other way, so a batch type that normally carries
+// debits only carries credits here, and its ServiceClassCode becomes
+// CreditsOnly. Validation that assumes the forward direction has to know about
+// this or it rejects a file the library itself produced.
+func (batch *Batch) IsReversal() bool {
+	if batch.Header == nil {
+		return false
+	}
+	return strings.EqualFold(strings.TrimSpace(batch.Header.CompanyEntryDescription), "REVERSAL")
+}
```
adamdecaf, line 305: Grok: `IsReversal` is sitting between `Create`'s comment and `Create` itself, so `Create` loses its godoc and this function's docs start with "Create will tabulate and assemble an ACH batch…".

Give `IsReversal` its own comment block (a one-liner is enough; see `IsADV`) and leave `Create`'s comment attached to `Create`.

Also, `batchTEL.go` still has the original inline `strings.EqualFold(...)` check. The PR description says that comparison is now `Batch.IsReversal`, but TEL wasn't updated — use the helper there too so the original copy can't drift.

adamdecaf (commented):
Grok review — not approving. A few things to change.

batchRCK.go
```diff
@@ -53,10 +53,14 @@ func (batch *BatchRCK) Validate() error {
...
-		return batch.Error("ServiceClassCode", ErrBatchServiceClassCode, batch.Header.ServiceClassCode)
+	// RCK detail entries can only be a debit, ServiceClassCode must allow debits.
+	// A reversal is the exception: it carries credits, so File.Reversal sets
+	// CreditsOnly here and this check would reject the library's own output.
+	if !batch.IsReversal() {
+		switch batch.Header.ServiceClassCode {
+		case CreditsOnly:
+			return batch.Error("ServiceClassCode", ErrBatchServiceClassCode, batch.Header.ServiceClassCode)
+		}
 	}
```
adamdecaf, line 64: Grok: Nacha already answers this in `reversal.go`: `REVERSAL` replaces the original description, including content otherwise required by the Rules. Skip the `REDEPCHECK` check below when `IsReversal()`, put RCK in `TestReversal_DebitOnlySECTypes`, and drop `TestReversal_RCKCannotValidate`.

As written, this credit/debit exception never runs — `Validate` still dies on `REDEPCHECK` first. Either finish that or leave `batchRCK.go` alone.

batchARC.go
```diff
@@ -62,10 +62,14 @@ func (batch *BatchARC) Validate() error {
...
-		return batch.Error("ServiceClassCode", ErrBatchServiceClassCode, batch.Header.ServiceClassCode)
+	// ARC detail entries can only be a debit, ServiceClassCode must allow debits.
+	// A reversal is the exception: it carries credits, so File.Reversal sets
+	// CreditsOnly here and this check would reject the library's own output.
+	if !batch.IsReversal() {
+		switch batch.Header.ServiceClassCode {
+		case CreditsOnly:
+			return batch.Error("ServiceClassCode", ErrBatchServiceClassCode, batch.Header.ServiceClassCode)
+		}
 	}
```
adamdecaf, line 73: Grok: These comments restate the PR body in every debit-only file. Match TEL:

```go
// Forward TEL batches can only have debit entries, but REVERSAL batches can only have credits
```

Same for BOC, POP, RCK, TRC, TRX, XCK.

The `if !batch.IsReversal() { switch … }` wrapper is also noisier than:

```go
if batch.Header.ServiceClassCode == CreditsOnly && !batch.IsReversal() {
    return batch.Error("ServiceClassCode", ErrBatchServiceClassCode, batch.Header.ServiceClassCode)
}
```

karpovantonme:
All addressed. RCK is finished rather than documented: `reversal.go` already quotes the rule that REVERSAL replaces the original description including content the Rules would otherwise require, so `Validate` skips the REDEPCHECK check on a reversal and RCK joins `TestReversal_DebitOnlySECTypes` -- the standalone test is gone. TEL calls the helper now so the two comparisons can't drift, `IsReversal` moved next to `IsADV` so `Create` keeps its godoc, and the per-file comments match the TEL wording. 71 packages pass

adamdecaf (approved):

PR #1857, opened 2026-09-14
ci: run hot-path Go benchmarks in this repository

adamdecaf:
## Why

ACH benches used to run in `moov-io/benchmarks`. The chart commit was a hash from that hub, not this repo.

## What

- `make bench` runs the existing hot-path suite (large-file read/write, merge, build, field parse/format). It does not `-bench=.` — most `Benchmark*` functions here are converted unit tests.
- Nightly (and `workflow_dispatch`) stores `docs/bench` labeled with the last non-result ACH commit.
- `push` CI ignores `docs/bench/**` so result commits do not re-run the test matrix.
- Pins, job-level permissions, and `persist-credentials: false` match the rest of moov-io CI.

The hub at https://moov-io.github.io/benchmarks/ will pick up `docs/bench/data.js` after the first run.

🤖 grok


adamdecaf:
Hub change that consumes `docs/bench/data.js` after this lands: https://github.com/moov-io/benchmarks/pull/18

🤖 grok

PR #1864, opened 2026-09-19
Fix batch validation to check AddendaRecordIndicator matches actual addenda count

### LINKED ISSUE #1647: Batch Validation Missed Addenda Format

jellyfishing2346:
## Summary
Fixes #1647

The batch validation was not catching when the `AddendaRecordIndicator` field doesn't match the actual number of addendas in an entry. This mismatch caused entries to be rejected by the Federal Reserve with return code R25.

## Changes
- Added validation in the `isAddendaSequence` function to ensure the `AddendaRecordIndicator` field matches the actual addenda count
- If `AddendaRecordIndicator=1`, validation now requires at least one addenda record
- If `AddendaRecordIndicator=0`, validation now requires no addenda records

## Testing
- Added test case for `AddendaRecordIndicator=1` with no addenda (validation fails as expected)
- Added test case for `AddendaRecordIndicator=0` with addenda present (validation fails as expected)
- Added test case for valid `AddendaRecordIndicator=1` with addenda (validation passes)
- Added test case for valid `AddendaRecordIndicator=0` with no addenda (validation passes)
- All existing batch validation tests pass
- All issue-specific tests pass


adamdecaf:
Pushed a follow-up for the `AddendaRecordIndicator=1` / no-addenda path:

- `ErrBatchAddendaRequired` (`is 1 but found no addenda record(s)`) instead of `ErrBatchExpectedAddendaCount`, which is the CTX/ATX/TRX count field
- Tests that used indicator `1` with no addenda as setup now expect this error, or drop the indicator so the original assertion still runs

🤖 grok

adamdecaf (approved):

PR #1866, opened 2026-09-21
fix: drop generated ADV batch control from prenote files

### LINKED ISSUE #1052: prenote file created with autogenerated ADV service class code
ACH Version: `v1.18.4`

From Zach,

> Another one here maybe - I just created a Prenote file and ended up with "serviceClassCode": 280, in the ADV Batch control section - mid you I did not make this section so this was completely generated by Moov ACH (edited) 

From Connor,
> pretty sure that documentation is incorrect. 280 is a service class code specifically for ADV batches

File created with the proper ServiceClassCodes and without errors.

jellyfishing2346:
Creating a prenote from JSON returned an ADV batch control with service class 280 that the caller did not send. Batch JSON decoding installed `NewADVBatchControl`, and that default stayed on non-ADV batches.

Decoding now keeps that control only when the payload includes `advBatchControl` or the batch is ADV. Building a non-ADV batch drops any ADV control, including one echoed from an older response. The batch control keeps the header service class (200, 220, or 225 for prenotes).

The ADV sample linked from the docs is accounting advice, so the example label now says Accounting Advice.

Fixes #1052


adamdecaf:
Thanks!

🤖 grok

